PR #2, opened 2020-11-09
Add integer compression/decompression.

samaloney:


stixcore/calibration/compression.py
```diff
@@ -0,0 +1,170 @@
...
+    >>> comp = compress(12345, s=0, k=5, m=3)
+    >>> int(comp)
+    92
+    >>> comp = compress(-1984, s=1, k=3, m=4)
+    >>> int(comp)
+    255
+    """
+    values = np.atleast_1d(np.array(values))
+    if not np.issubdtype(values.dtype, np.integer):
+        raise ValueError(f'Input must be an integer type not {values.dtype}')
```
nicHoch, line 65: we could create your own error types like: NonIntCompressionError

nicHoch (approved):
we also discussed the usage of default parameters (s, k m ..).
but came to the conclusion that the codebase should have as less implicit states as possible

PR #3, opened 2020-11-10
Inital spice manager for time conversion

samaloney:


stixcore/ephemeris/manager.py
```diff
@@ -0,0 +1,111 @@
...
+        Parameters
+        ----------
+        meta_kernel_path : `str` or `pathlib.Path`
+            Path to the meta kernel
+
+        """
+        self.mk_path = Path(meta_kernel_path)
+        *_, datestamp, version = self.mk_path.name.split('_')
+        self.kernel_date = datetime.strptime(datestamp, '%Y%m%d')
+        spice.furnsh(str(meta_kernel_path))
```
samaloney, line 29: I think as spicepy is a wrapper around the CSPICE library it appears to be global in that kernels loaded in one module / function will be available anywhere spicepy is imported. A actual context manager might be a better approach but not sure.

stixcore/ephemeris/manager.py
```diff
@@ -0,0 +1,111 @@
...
+            Time to convert to SCET
+
+        Returns
+        -------
+        `str`
+            SCET of datetime
+        """
+        if isinstance(datetime, Time):
+            datetime = datetime.to_datetime()
+        et = spice.datetime2et(datetime)
```
nicHoch, line 109: a type check might help in long term 
crashes with: datetime_to_scet(1)

nicHoch (approved):
would suggest some more type checks in the time conversion functions.

i am not sure how it will resolve: in https://github.com/i4Ds/STIXCore/pull/2

all the example_module files are already gone here they axist again.

stixcore/ephemeris/manager.py
samaloney: I can add a check to make sure its a string, it's not very pythonic in general but in this case probably makes sense.

samaloney:
> would suggest some more type checks in the time conversion functions.
> 
> i am not sure how it will resolve: in #2
> 
> all the example_module files are already gone here they axist again.

Yea once #2 is merged this PR will become invalid and I'll have to update to match master so will pick up these changes

PR #4, opened 2020-11-11
TMTC parsing with demo for TM(1,1) and TM(21, 6) SSID=30

samaloney:
Uses a two level factory pattern to separate packet creating logic from packet classed themselves.

Currently a demo is hard coded for TM(1,1) and TM(21, 6) SSID=30 need to interface with IDB to create programatically


stixcore/tmtc/packets.py
```diff
@@ -0,0 +1,317 @@
...
+        return self.__repr__()
+
+
+class TM_1_1(GenericTMPacket):
+    """
+    TM(1,1) Telecommand acceptance report
+    """
+    def __init__(self, data):
+        super().__init__(data)
+        structure = {'NIX00001': 'uint:16', 'NIX00002': 'uint:16'}
```
samaloney, line 260: @drhlxiao would your IDB routines be able to give back the packet structure based on service type, service subtype and SSID?

stixcore/tmtc/packets.py
```diff
@@ -0,0 +1,317 @@
...
+                                  'NIX00272': ', '.join(['uint:8']*num_data_points)}, num_energies),
+                                ({'NIX00273': 'uint:16'}, 1),
+                                ({'NIX00274': ', '.join(['uint:8']*num_data_points)}, 1),
+                                ({'NIX00275': 'uint:16'}, 1),
+                                ({'NIX00276': ', '.join(['uint:8']*num_data_points)}, 1)]
+
+        for i, (structure, repeats) in enumerate(repeaters_structures):
+            dynamic = parse_repeated(self.source_packet_header.bitstream, structure, repeats)
+            attr_dict.update(dynamic['fields'])
+
```
samaloney, line 311: @drhlxiao same kind of question I've hard coded this example but ideally would like to be able to query the IDB using your routines to obtain the fixed and dynamic structure based on the service type, subtype and SSID

samaloney: @nicHoch I think it would be good for you to review this part and we can merge and then I can review the other part in #19. What do you think?

samaloney:
This code was integrated in #19

PR #6, opened 2020-11-19
Add routines for spice calculations for positions, orientation and conversion to instrument frame.

samaloney:


stixcore/data/test.py
```diff
@@ -0,0 +1,31 @@
...
+TEST_DATA_NAME = 'test_data.tar.gz'
+
+
+def _download_test_tar(overwrite=False):
+    destination = data_dir / TEST_DATA_NAME
+    if destination.exists() and overwrite is False:
+        return
+    with urlopen(TEST_DATA_BASE_URL+TEST_DATA_NAME) as req:
+        with destination.open(mode='wb') as file:
+            file.write(req.read())
```
nicHoch, line 20: maybe use:

urllib.request.urlretrieve

stixcore/ephemeris/manager.py
```diff
@@ -89,29 +133,115 @@ def scet_to_datetime(scet):
...
+class Position(SpiceManager):
+    """
+    Obtain spacecraft position and orientation, convert to and from instrument coordinate system
+    """
+    spice_context = SpiceManager.spice_context
+
+    @spice_context
+    def get_position(self, *, date, frame):
+        """
+        Get the position of SolarOrbiter.
```
nicHoch, line 174: Get the position of SolarOrbiter **at a given time and coordinate frame.** 

Is this clear because of the parameters

nicHoch (approved):

stixcore/data/test.py
samaloney: `urllib.request.urlretrieve` considered legacy maybe removed in the future https://docs.python.org/3.6/library/urllib.request.html#legacy-interface

PR #19, opened 2020-12-07
TM_1_1 idb test

nicHoch:


stixcore/tmtc/packet_factory.py
```diff
@@ -0,0 +1,115 @@
...
+        packet = self._check_registered(sph, idb)
+        return self.tm_packet_factory(packet, idb)
+
+
+class TMPacketFactory(BaseFactory):
+    """
+    Factory from TM packet return the correct type of based on the packet data and registered.
+    """
+    def __call__(self, data, idb):
+        return self._check_registered(data, idb)
```
samaloney, line 93: We'll have to think about this - when and how often should the IDB be queried. In theory every packet could need a different version of the IDB so maybe we need to create and instance of the `IDBManager` and and for each packet check if the current version is applicable or not?

stixcore/idb/idb.py
```diff
@@ -572,3 +995,142 @@ def get_idb_version(self):
...
+        args = (service_type, service_subtype)
+        if ssid is not None:
+            args = args + (ssid,)
+        param_pcf_structures = self._execute(sql, args, 'dict')
+
+        repeater = [{'node': IdbPacketTree(), 'counter': 1024}]
+
+        for par in param_pcf_structures:
+            parObj = IdbVariableParameter(par)
+            if repeater:
```
samaloney, line 1122: with `repeater` defined as above is this `if` not always true?

stixcore/tmtc/parser.py
```diff
@@ -0,0 +1,173 @@
...
+from stixcore.util.logging import get_logger
+
+logger = get_logger(__name__)
+
+
+class PacketData:
+    """Generic class for organizing all parameters of the TM packet data."""
+
+
+def parameter_dict_2_PacketData(d):
```
samaloney, line 14: Maybe make this a class method on the `PacketData` class?

samaloney (commented):
Looks good, the reason some of the functions/classes are not showing up in the docs is they have not been added to the `__all__ = `  at the top of the relevant file.

stixcore/idb/idb.py
```diff
@@ -11,11 +11,436 @@
 logger = get_logger(__name__)
 
 
+class IdbData:
```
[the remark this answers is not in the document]
nicHoch: good point i added the new super class

stixcore/idb/idb.py
nicHoch: repeater.pop() can empty the repeater list and so this statement will become false

stixcore/tmtc/parser.py
nicHoch: good point

stixcore/tmtc/packet_factory.py
nicHoch: made an issue for that

stixcore/idb/manager.py
```diff
@@ -26,12 +25,11 @@
 logger = get_logger(__name__)
 
 
-class IDBManager:
-    """Manages IDB (definition of TM/TC packet structures) Versions
-    and provides a IDB reader
-    """
+class IdbManager:
```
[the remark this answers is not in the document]
nicHoch: grr i renamed that yesterday as i thought you suggested IdbManager.
Will change again

samaloney (approved):

PR #23, opened 2020-12-11
Add calculation of the absortion of X-ray by non-imaging components

samaloney:
I'm not sure what to do with this it not great, I have some ideas to improve it as but it does what we need right now, might wait until a need arise.

- The material and components are fixed in the code should probably be read the configuration database to allow different versions
- The detector are all treated the same the fine grid cover is a bit add-hoc could move to per detector approach
- Hard to really test without using the output values form the code itself ...

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -18,6 +18,7 @@ install_requires =
     sunpy
     spiceypy
     bitstring
+    roentgen
 
 
 [options.entry_points]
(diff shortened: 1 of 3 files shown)


nicHoch (approved):
this look all good to my beside my personal preferences for spaces in formulas. 
for the math behind all that i can not really tell.

PR #22, opened 2020-12-08
Add live time fraction calculation

samaloney:

PR #28, opened 2020-12-18
fix #22: add ref to IdbManager to parsing

nicHoch:
i think your pattern : 

```
idb_mnager = IDBManger(...)
....
idb = IDBManger.get_version(date)
....
idb.get_packet_type_info(...)
```

is fine and does not need a lot of refactoring and coding.
i implemented this.

DateTime: is just s stub and very work in progress.


samaloney (approved):
Some small suggestions but looks good to go.

PR #31, opened 2020-12-18
Refactor packet foractory to handle IDB

samaloney:


samaloney:
@nicHoch what do you think, it's not perfect but a little bit cleaner

stixcore/tmtc/packets.py
```diff
@@ -214,6 +183,7 @@ class GenericPacket:
 
     """
     _registry = dict()
+    idb_manger = None
```
nicHoch, line 186: typo or common abbreviation?

nicHoch (approved):
all good

stixcore/tmtc/packets.py
samaloney: typo for sure

PR #34, opened 2020-12-22
Feat make spice kernel from SOC location indpendent

### LINKED ISSUE #11: Make spicemanger work independent of kernel location.
Meta kernels distributed from the SOC contain a `PATH_VALUES` parameter and in oder for the kernels to work properly either you have to change the working directory to match the path of the kernels or change the `PATH_VALUES` parameter to match the kernel location.

> Implementation Notes
> -------------------------------------------------------------------------
> 
>    It is recommended that users make a local copy of this file and
>    modify the value of the PATH_VALUES keyword to point to the actual
>    location of the Solar_Orbiter SPICE data set's ``data'' directory on
>    their system. Replacing ``/'' with ``\'' and converting line
>    terminators to the format native to the user's system may also be
>    required if this meta-kernel is to be used on a non-UNIX workstation.

The idea would be to make a temporary copy of the meta kernel updating the `PATH_VALUES` variable in this copy to point the real location. Below is an example of the start of a meta kernel file so `PATH_VALUES` should be updated to match the parent directory of the given meta kernel.

```
   \begindata

     PATH_VALUES       = ( '..' )

     PATH_SYMBOLS      = ( 'KERNELS' )

     KERNELS_TO_LOAD   = (

                           '$KERNELS/ck/solo_ANC_soc-sc-iboom-ck_20180930-21000101_V01.bc'
                           '$KERNELS/ck/solo_ANC_soc-sc-oboom-ck_20180930-21000101_V01.bc'
                           '$KERNELS/ck/solo_ANC_soc-sc-fof-ck_20180930-21000101_V02.bc'
                           ...
                            )
```

samaloney:
Create a temporary copy and update the path variable to match the location of the kernels close #11

### DIFF
--- a/stixcore/ephemeris/manager.py
+++ b/stixcore/ephemeris/manager.py
@@ -1,9 +1,11 @@
 import os
 import re
+import logging
 import tempfile
 import functools
 from pathlib import Path
 from datetime import datetime
+from textwrap import wrap
 
 import numpy as np
 import spiceypy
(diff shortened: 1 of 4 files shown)


nicHoch (approved):
looks good

PR #39, opened 2021-01-08
add decompression to packet processing

nicHoch:
work in progress!

### DIFF
--- a/stixcore/processing/decompression.py
+++ b/stixcore/processing/decompression.py
@@ -1,0 +1,15 @@
+from stixcore.calibration.compression import decompress as algo_decompress
+
+
+def work_decompress(val, skm):
+    return algo_decompress(val, s=skm[0], k=skm[1], m=skm[2])
+
+
+def decompress(packet):
+    params = packet.get_decompression_parameter()
+    if not params:
+        return
+    for param_name, (sn, kn, mn) in params.items():
+        skm = (packet.data.get_first(sn), packet.data.get_first(kn), packet.data.get_first(mn))
+        c = packet.data.work(param_name, work_decompress, skm)
+        print(c, param_name, work_decompress, skm)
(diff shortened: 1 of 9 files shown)


samaloney:
I wonder if we shouldn't flatten the packets as an intermediate step, it would avoid all that set_path, get_path, get_all etc what do you think?

stixcore/processing/decompression.py
```diff
@@ -0,0 +1,18 @@
...
+def decompress(packet):
+    decompression_parameter = packet.get_decompression_parameter()
+    if not decompression_parameter:
+        return 0
+    c = 1
+    for param_name, (sn, kn, mn) in decompression_parameter.items():
+        skm = (sn if isinstance(sn, int) else packet.data.get(sn),
+               packet.data.get(kn),
+               packet.data.get(mn))
+        c += packet.data.work(param_name, work_decompress, skm)
```
nicHoch, line 17: @samaloney you might wonder about that. your decompression method could deal with all all repeated NIX in the same packet at once. but then in generates a regular (n,m) array as output what might fail if there is no regular m.

so i process each array in the list as separate call within the work method. could also be helpfull for other processing steps later.

stixcore/tmtc/parser.py
```diff
@@ -37,10 +38,130 @@ def parameter_dict_2_PacketData(cls, d):
...
+            The NIX name of the parameter.
+
+        Returns
+        -------
+        'any'
+            The found value.
+        """
+        return getattr(self, nix)
+
+    def apply(self, nix, callback, args, addnix=None):
```
samaloney, line 132: you could add **kwargs here and pass them to the callback too if we need to might be useful later

stixcore/tmtc/parser.py
```diff
@@ -37,10 +38,130 @@ def parameter_dict_2_PacketData(cls, d):
...
+        """Apply a processing method to a parameter.
+
+        Overids the value after processing or creates a new one.
+
+        Parameters
+        ----------
+         nix : `str`
+            The NIX name of the parameter.
+        callback : function
```
samaloney, line 141: Could be a class either - 'callable' might be more descriptive

samaloney (approved):
Looks good to me - a few comment but nothing that _has_ to be changed.

PR #41, opened 2021-01-12
Add a high level datetime object to handle conversion from SCET/OBT to UTC

samaloney:


stixcore/datetime/datetime.py
```diff
@@ -0,0 +1,58 @@
...
+        Parameters
+        ----------
+        coarse : `int`
+            Coarse time stamp (seconds)
+        fine : `int`
+            Fine time stamp fraction of seconds 1/2**31 -1
+        """
+        # Convention if top bit is set means times are not synchronised
+        self.time_sync = (coarse >> 31) != 1
+        self.coarse = coarse if self.time_sync else coarse ^ 2**31
```
nicHoch, line 38: as operator precedence is not always intuitive i like brackets

stixcore/datetime/datetime.py
```diff
@@ -0,0 +1,58 @@
...
+        -------
+        `datetime.datetime`
+            The corresponding UTC datetime object.
+        """
+        with SPICE_TIME as time:
+            utc = time.scet_to_datetime(f'{self.coarse}:{self.fine}')
+            return utc
+
+    def __repr__(self):
+        return f'{self.__class__.__name__}(coarse={self.coarse}, fine={self.fine})'
```
nicHoch, line 55: self.__class__.__name__ good point

nicHoch (approved):
nice

PR #52, opened 2021-01-21
LevelB (binary processing) port of code from STIX-python-parser

samaloney:
Just to have a working version to build from/talk about.


nicHoch:
removed for #53

PR #53, opened 2021-01-22
Moc file and directory handling + LevelB Processing

nicHoch:


stixcore/soc/manager.py
```diff
@@ -0,0 +1,86 @@
...
+        if not self.file.exists():
+            raise ValueError(f"file path not found: {file}")
+        stat = self.file.stat()
+        self.size = stat.st_size
+        self.file_date = stat.st_ctime
+        self.spacecraft_date = None  # TODO some file name conventions?
+        self.tmtc = TMTC.All
+        self.packet_data = defaultdict(list)
+
+        with open(self.file, "r") as f:
```
samaloney, line 28: `with self.file.open('r') as f`

samaloney: can search for tags in the xml file e.g `tree.findall('*PktRawResponse')`

stixcore/soc/manager.py
```diff
@@ -0,0 +1,86 @@
...
+            if 'PktRawResponse' in header:
+                self.tmtc = TMTC.TM
+            elif 'PktTcReportResponse' in header:
+                self.tmtc = TMTC.TC
+            else:
+                raise ValueError("xml does not contain any TM or TC response")
+
+    def read_packet_data_header(self):
+        root = Et.parse(str(self.file)).getroot()
+        if self.tmtc == TMTC.TC:
```
samaloney, line 42: I would hope both these beaches can be combined using TMTCPacket() but might take some work.

stixcore/soc/manager.py
```diff
@@ -0,0 +1,86 @@
...
+    def get_files(self, tmtc=TMTC.All):
+        files = list()
+
+        filter = "*.xml"
+        if tmtc == TMTC.TC:
+            filter = "*PktTcReport*.xml"
+        elif tmtc == TMTC.TM:
+            filter = "*PktTmRaw*.xml"
+
+        for filename in sorted(list(self.data_root.glob(filter)), key=os.path.getmtime):
```
samaloney, line 79: I can't remember which but need to becareful with `ctime` and `mtime` as they are not same on window/linux

stixcore/soc/manager.py
```diff
@@ -0,0 +1,86 @@
...
+
+class SOCPacketFile:
+    def __init__(self, file):
+        self.file = Path(file)
+        if not self.file.exists():
+            raise ValueError(f"file path not found: {file}")
+        stat = self.file.stat()
+        self.size = stat.st_size
+        self.file_date = stat.st_ctime
+        self.spacecraft_date = None  # TODO some file name conventions?
```
samaloney, line 24: Yea so this is related to the email to Hualin about trying to standardise the filenames they are some what standard now
```
LTP03_20200101_manual_BatchRequest.PktTmRaw.SOL.0.2021.001.11.03.21.142.SRHD@2021.001.11.03.22.268.1.xml
<              name              >.< type     >.<      req date            >@<data date           >.<part>.xml
```

samaloney (commented):
No issue with this but need to think about how it will fit together

stixcore/soc/manager.py
nicHoch: yes but this takes all or nothing. annd the xml parsing needs quite some time. here at this stage it is just some "security" layer - to check if the content of the xml is "somehow" ok (as expected)
the entire file is not processed at this stage

stixcore/io/soc/manager.py
```diff
@@ -0,0 +1,89 @@
...
+        root = Et.parse(str(self.file)).getroot()
+        if self.tmtc == TMTC.TC:
+            for i, node in enumerate(root.iter('PktTcReportListElement')):
+                # TODO add TC packer reading
+                pass
+        elif self.tmtc == TMTC.TM:
+            for i, node in enumerate(root.iter('Packet')):
+                packet_binary = unhexlify(node.text)
+                # Not sure why guess and extra moc header
+                tm_packet = TMPacket(packet_binary[76:])
```
samaloney, line 53: I'm wondering if the actual Parsing doesn't belong in the `processing/leveb/LevelB` class - better separation of concerns?

stixcore/io/soc/manager.py
```diff
@@ -0,0 +1,89 @@
...
+                raise ValueError("xml does not contain any 'Response'")
+            if 'PktRawResponse' in header:
+                self.tmtc = TMTC.TM
+            elif 'PktTcReportResponse' in header:
+                self.tmtc = TMTC.TC
+            else:
+                raise ValueError("xml does not contain any TM or TC response")
+
+    def free_packet_data(self):
+        self.packet_data = defaultdict(list)
```
samaloney, line 41: I suppose what I was tying to say in the chat was with actually checking the memory profile we probably shouldn't do this or `del self.packet`. Why are we keeping a copy of the parsed packets anyway whey not return them from the `read_packet_data_header` call?

stixcore/processing/levelb.py
```diff
@@ -0,0 +1,150 @@
...
+        self.level = 'LB'
+        self.type = f"{control['service_type'][0]}-{control['service_subtype'][0]}"
+        self.control = control
+        self.data = data
+
+        synced_times = (self.control['scet_coarse'] >> 31) != 1
+        self.control['scet_coarse'] = np.where(synced_times, self.control['scet_coarse'],
+                                               self.control['scet_coarse'] ^ 2**31)
+        self.control['time_sync'] = synced_times
+        scet = self.control['scet_coarse'] + self.control['scet_coarse']/2**16
```
samaloney, line 36: Can probably replace some of this logic with our DateTime class but this can be done in a separate issue commenting here before I forget

stixcore/processing/levelb.py
```diff
@@ -0,0 +1,150 @@
...
+import logging
+from time import perf_counter
+from pathlib import Path
+
+import numpy as np
+
+from astropy.table import unique, vstack
+from astropy.table.table import QTable, Table
+
+from stixcore.io.fits.processors import SEC_IN_DAY, FitsLBProcessor
```
samaloney, line 11: Can probably replace this with astroy units `(1 * u.d).to_value('s')` but not sure if we should

stixcore/processing/levelb.py
```diff
@@ -0,0 +1,150 @@
...
+            data['control_index'] = data['control_index'] - min_index
+
+            yield type(self)(control=control, data=data)
+
+    @classmethod
+    def process(cls, tmfile, fits_processor):
+        tmfile.read_packet_data_header()
+
+        for product, packets in tmfile.packet_data.items():
+
```
samaloney, line 110: So I think the packet parsing would be better here and in side a try except where we can log the info of the packer if the parsing fails.

stixcore/io/fits/processors.py
```diff
@@ -0,0 +1,140 @@
...
+
+class FitsLBProcessor(FitsProcessor):
+    def __init__(self, archive_path):
+        self.archive_path = archive_path
+
+    def write_fits(self, product):
+
+        for prod in product.to_days():
+            filename = self.generate_filename(prod, version=1)
+            if getattr(prod, 'ssid', False):
```
samaloney, line 31: Not sure if it will ever have `ssid` with the current code?

samaloney (commented):
This is on the right track - once we have L0 and L1 we probably need to do a through review and refactor.

stixcore/io/soc/manager.py
nicHoch: moved

stixcore/processing/levelb.py
nicHoch: open for now

stixcore/processing/levelb.py
nicHoch: done

stixcore/io/fits/processors.py
nicHoch: fixed type now also includes a ssis/sp1_val

stixcore/processing/levelb.py
nicHoch: not done at the moment

stixcore/io/soc/manager.py
nicHoch: replaced with a generator pattern

samaloney (approved):

PR #70, opened 2021-02-11
add energy channel reading from config

nicHoch:


samaloney (approved):
I think the read the doc config need to be configured to include the submodule too.

PR #72, opened 2021-02-12
add decompression

nicHoch:


nicHoch:
moved to https://github.com/i4Ds/STIXCore/pull/80

PR #75, opened 2021-02-21
Refactor parameters/PacketData to represent repeated NIX values as lists

samaloney:


nicHoch (approved):
very good cleanup.
nothing to complain in detail. Looking forward to work with this clean interface

stixcore/tmtc/tests/test_packet_factory.py
```diff
@@ -339,6 +339,7 @@ def test_engineering(data_dir, idbm, packets):
         assert c == 0
 
 
+@pytest.mark.skip(reason='Broken on py3.6')
```
samaloney, line 342: I had to skip this as it was killing the CI it was working locally but need to update CI to py37 anyway so will see if it work on 3.7+

PR #76, opened 2021-02-26
add lfs support for (test) data

nicHoch:


stixcore/ephemeris/manager.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, meta_kernel_path):
 
         with self.meta_kernel_path.open('r') as mk:
             original_mk = mk.read()
-            kernel_dir = str(self.meta_kernel_path.parent.parent.resolve())
+            kernel_dir = str(self.meta_kernel_path.parent.resolve())
```
samaloney, line 50: hopefully we should be able to replace all these with the imported paths so something like:
```python
from stixcore.data import test
...
kernel_dir  = str(test.ephemeris.META_KERNEL)
```
maybe for another issue?

samaloney (changes requested):
So many files! Think it good apart from the RTD fix for LFS

samaloney:
I tried to push a fix for the docs but failed with authentication error - do you remerge if you checked the allow edits from maintainers box? Any way if you pull from upstream master and rebase it should be ok.

docs/conf.py
```diff
@@ -83,6 +83,15 @@
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 
+# Workaround to install and execute git-lfs on Read the Docs
+import os
+if not os.path.exists('./git-lfs'):
```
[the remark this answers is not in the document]
nicHoch: add an aditional check

stixcore/ephemeris/manager.py
nicHoch: i changed that

docs/conf.py
```diff
@@ -106,6 +106,23 @@
...
+    os.system('tar xvfz git-lfs-linux-amd64-v2.13.2.tar.gz')
+    os.system('./git-lfs install')  # make lfs available in current repository
+    os.system('./git-lfs fetch')  # download content from remote
+    os.system('./git-lfs checkout')  # make local files to have the real content on them
+    import time
+    os.system('./git-lfs logs last')  # echo test
+    # os.system('git lfs install')  # report any errors
+    # os.system('echo "git pull after lfs"')  # echo test
+    # os.system('git checkout master')  # report any errors
+    # os.system('git pull')  # report any errors
```
samaloney, line 123: Delete maybe?

setup.cfg
```diff
@@ -31,6 +31,7 @@ test =
     pytest
     pytest-astropy
     pytest-cov
+    git_lfs
```
samaloney, line 34: Is this needed for RTD, I don't see it imported in any of the modules?

samaloney (approved):

samaloney:
This should be squash merged

PR #79, opened 2021-03-03
Update transmission generation code

samaloney:
* Evalute tranmssion at energies not centers
* Add energies to the returned tables
* Update tests for theses changes
* Add function to write transmission tables to csv


nicHoch (approved):
technically it looks all good

PR #80, opened 2021-03-03
add decompression to QL LC and Compressed/SummedPixelData

nicHoch:


stixcore/products/tests/test_bsd.py
```diff
@@ -0,0 +1,27 @@
...
+@patch('stixcore.products.levelb.binary.LevelB')
+@pytest.mark.parametrize('packets', testpackets, ids=[f[0].name for f in testpackets])
+def test_xray(levelb, packets):
+    hex_file, cl, name = packets
+    with hex_file.open('r') as file:
+        hex = file.readlines()
+
+    levelb.data.__getitem__.return_value = [re.sub(r"\s+", "", h) for h in hex]
+
+    xray_L0 = cl.from_levelb(levelb)
```
samaloney, line 21: So this bypassed the code which generates and check for a complete packet sequence passing it incomplete sequence will give undefined behaviour. I'm not sure where is best to put this kind of check I think maybe we can expand and subclass `PacketSequence` or add `CompletePacketSequence` class?

stixcore/products/product.py
```diff
@@ -38,6 +42,18 @@ def __init_subclass__(cls, **kwargs):
         if hasattr(cls, 'is_datasource_for'):
             cls._registry[cls] = cls.is_datasource_for
 
+    @classmethod
+    def from_levelb(cls, levelb):
+
+        packets = [Packet(unhexlify(d)) for d in levelb.data['data']]
+
+        for packet in packets:
+            decompression.decompress(packet)
```
samaloney, line 51: This is actually a better approach as its possible that compression or calibration parameters could change between packets.

stixcore/products/level0/science.py
```diff
@@ -124,14 +124,11 @@ def from_levelb(cls, levelb):
...
-            nix = f'NIX00{i}'
-            triggers.extend(packets.get_value(nix))
-            triggers_var.extend(packets.get_value(nix, attr='error'))
-
-        data['triggers'] = np.array(triggers).T
+        triggers = np.array([packets.get_value(f'NIX00{i}') for i in range(242, 258)])
+        triggers_var = np.array([packets.get_value(f'NIX00{i}', attr='error')
+                                 for i in range(242, 258)])
+
+        data['triggers'] = triggers.T
```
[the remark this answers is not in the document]
samaloney: Bit of both, list comprehension are usually faster than the equivalent loop e.g.
```ptyhon
[a**2 for a in list]
```
is usually faster then
```python
r = []
for a in list:
    r.append(a**2)
```
but maybe not twice as fast. I'll see if I can get the other single loop to work.

PR #87, opened 2021-03-15
Feat visibility bsd

samaloney:


stixcore/products/level0/quicklook.py
```diff
@@ -148,15 +148,15 @@ def from_levelb(cls, levelb):
                                      control[['num_samples', 'index']]])
 
         control['compression_scheme_counts_skm'], control['compression_scheme_counts_skm'].meta =\
-            _get_compression_scheme(packets, 'NIX00272')
+            _get_compression_scheme(packets, ['NIXD0101', 'NIXD0102', 'NIXD0103'])
```
nicHoch, line 151: i would prefer the "old" approach so the code do not have to know hard coded which are the nix parametrs used for decompression. As this might be false as the actually uncompression takes the nix names from somewhere else.

stixcore/products/level0/science.py
```diff
@@ -266,14 +266,253 @@ def is_datasource_for(cls, *, service_type, service_subtype, ssid, **kwargs):
...
+        # vis[:, detector_id[0], e_low.reshape(unique_times.size, -1)[0]] = (
+        #         np.array(packets['NIX00263']) + np.array(packets['NIX00264'])
+        #         * 1j).reshape(unique_times.size, num_detectors, -1)
+
+        real = packets.get_value('NIX00263')
+        real_var = packets.get_value('NIX00263', attr='error')
+        imaginary = packets.get_value('NIX00264')
+        imaginary_var = packets.get_value('NIX00264', attr='error')
+
+        mesh = np.ix_(np.arange(unique_times.size), detector_id[0][0],
```
nicHoch, line 371: we should store the skm values used for the decompression:

`control['compression_scheme_vis_skm'], \     control['compression_scheme_vis_skm'].meta = \    _get_compression_scheme(packets, 'NIX00263')`

stixcore/products/level0/science.py
```diff
@@ -266,14 +266,253 @@ def is_datasource_for(cls, *, service_type, service_subtype, ssid, **kwargs):
...
+
+        triggers = packets.get_value('NIX00267')
+        triggers_var = packets.get_value('NIX00267', attr='error')
+
+        data['triggers'] = triggers
+        data['triggers_err'] = np.sqrt(triggers_var)
+        data['counts'] = full_counts * u.ct
+        data['counts_err'] = np.sqrt(full_counts_var) * u.ct
+        data['control_index'] = 0
+
```
nicHoch, line 511: we should store the skm values used for the decompression:

`control['compression_scheme_triggers_skm'], \     control['compression_scheme_vis_skm'].meta = \    _get_compression_scheme(packets, 'NIX00267')`  also for counts

stixcore/products/product.py
```diff
@@ -213,11 +213,11 @@ def from_packets(cls, packets):
...
             control['compression_scheme_counts_skm'].meta = \
-            _get_compression_scheme(packets, 'NIX00260')
+            _get_compression_scheme(packets, ['NIXD0007', 'NIXD0008', 'NIXD0009'])
 
         control['compression_scheme_triggers_skm'], \
             control['compression_scheme_triggers_skm'].meta = \
-            _get_compression_scheme(packets, 'NIX00242')
+            _get_compression_scheme(packets, ['NIXD0010', 'NIXD0011', 'NIXD0012'])
 
         control['time_stamp'] = np.array(packets.get_value('NIX00402'))
```
nicHoch, line 222: remove that hard coded skm values. as the nixes are not constant and also not all packets have this information (visibilities->counts)

nicHoch (changes requested):
i looked mainly on the decompression part. and i think this should be reworked.
but the actual Visibilitie and Spectrogram logic looks good

samaloney:
I reworked (nearly identical to before) `_get_compression_scheme` and pull the calls out from the common `Control.from_packets` method and added them to the specific products updating the NIX values.

nicHoch (approved):
looks good now

PR #89, opened 2021-03-16
Add L0 housekeeping products

samaloney:
* Mini report TM(3, 25), SSID 1
* Maxi report TM(3, 25), SSID 2

### DIFF
--- a/stixcore/products/level0/housekeeping.py
+++ b/stixcore/products/level0/housekeeping.py
@@ -1,0 +1,204 @@
+"""
+House Keeping data products
+"""
+from stixcore.datetime.datetime import DateTime
+from stixcore.products.level0.quicklook import QLProduct
+from stixcore.products.product import Control, Data
+
+__all__ = ['MiniReport', 'MaxiReport']
+
+
+class MiniReport(QLProduct):
+    """
+    Mini house keeping reported during start up of the flight software.
+    """
+    def __init__(self, *, service_type, service_subtype, ssid, control, data, **kwargs):
+        super().__init__(service_type=service_type, service_subtype=service_subtype,
+                         ssid=ssid, control=control, data=data, **kwargs)
+        self.name = 'mini'
+        self.level = 'L0'
+        self.type = 'hk'
+
+    @classmethod
+    def from_levelb(cls, levelb):
+        packets = super().from_levelb(levelb)
+
+        service_type = packets.get('service_type')[0]
+        service_subtype = packets.get('service_subtype')[0]
+        ssid = packets.get('pi1_val')[0]
+
+        control = Control()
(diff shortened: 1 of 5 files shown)


stixcore/products/level0/housekeeping.py
```diff
@@ -0,0 +1,204 @@
...
+        control['index'] = range(len(control))
+
+        # Create array of times as dt from date_obs
+        times = [DateTime(ct, ft).as_float() for ct, ft in control['scet_coarse', 'scet_fine']]
+
+        # Data
+        data = Data()
+        data['time'] = times
+        data['sw_running'] = packets.get_value('NIXD0021', attr='value')
+        data['instrument_number'] = packets.get_value('NIXD0022', attr='value')
```
samaloney, line 120: Could probably replace this and the similar block above with a loop using a modified version of the the parameters name from the IDB? Override one that need to be modified or fix in idb

stixcore/products/level0/housekeeping.py
```diff
@@ -0,0 +1,204 @@
...
+        data['instrument_mode'] = packets.get_value('NIXD0023', attr='value')
+        data['hk_dpu_pcb_t'] = packets.get_value('NIXD0025')
+        data['hk_dpu_fpga_t'] = packets.get_value('NIXD0026')
+        data['hk_dpu_3v3_c'] = packets.get_value('NIXD0027')
+        data['hk_dpu_2v5_c'] = packets.get_value('NIXD0028')
+        data['hk_dpu_1v5_c'] = packets.get_value('NIXD0029')
+        data['hk_dpu_spw_c'] = packets.get_value('NIXD0030')
+        data['hk_dpu_spw0_v'] = packets.get_value('NIXD0031')
+        data['hk_dpu_spw1_v'] = packets.get_value('NIXD0032')
+        data['sw_version'] = packets.get_value('NIXD0001')
```
nicHoch, line 53: i am not sure but is it possible that this values here are already converted into engineering values?

so fare we only used times and this is ok for level0.
but my understanding was that we will for this voltages.... store the the raw value in level0 and later on in level1 convert it to engineering

stixcore/products/level0/housekeeping.py
nicHoch: hm yes that might be a good sugestion

nicHoch (approved):
did not really check for completeness of the nixes as i think the suggested generic "loop" over all HK params is a good approach.

stixcore/products/level0/housekeeping.py
samaloney: Yea so with the current setup they are converted here already but as you note this isn't the plan

PR #104, opened 2021-04-15
Fix small bug generating incorrect aspect duration.

samaloney:
When doing a final review on the aspect fits files I realised there was a small error in the calculation of the durations.


nicHoch (approved):
yes looks like not feasible for the IDB framework

PR #106, opened 2021-04-16
L1 fits write

nicHoch:


stixcore/products/level0/quicklook.py
```diff
@@ -49,11 +49,14 @@ def __init__(self, *, service_type, service_subtype, ssid, control, data,
...
-                                          + self.control['integration_time'][-1] / 2)
-        self.obs_avg = self.obs_beg + (self.obs_end - self.obs_beg) / 2
+        try:
+            self.obs_beg = SCETime.from_float(self.data['time'][0]
+                                              - self.control['integration_time'][0] / 2)
+            self.obs_end = SCETime.from_float(self.data['time'][-1]
+                                              + self.control['integration_time'][-1] / 2)
+            self.obs_avg = self.obs_beg + (self.obs_end - self.obs_beg) / 2
+        except ValueError:
+            pass
```
samaloney, line 59: Hum if we get a value error here I think we are in bad shape do you know what was causing it?

stixcore/products/level1/quicklook.py
```diff
@@ -1,142 +1,37 @@
...
-        self.service_subtype = service_subtype
-        self.ssid = ssid
-        self.control = control
-        self.data = data
-
-        energies = kwargs.get('energies', False)
-        if not energies:
-            energies = self.get_energies()
-        self.energies = energies
+class QLProduct(QLProductL0):
```
samaloney, line 18: it's a bit odd to have the baseclass like this maybe we should rename, but not entirely sure why this was done so maybe it's fine?

samaloney (approved):
Looks good to me. We now have a lot of references to idbs being passed around are they actual idb objects or the version labels?

stixcore/products/level0/quicklook.py
nicHoch: yes this was about my question about the timing conversions
some old test fits file had SCETime other "real" time within

stixcore/products/level1/quicklook.py
nicHoch: the QLProductL0 is just an import rename. Basically it means that this Level1QLProduct inherits from L0QLProduct

stixcore/processing/engineering.py
```diff
@@ -168,8 +168,12 @@ def raw_to_engineering_product(product, idbm):
 
         # replace the old column with the converted
         product.data[col] = product.data[CCN]
+        product.data[col].meta = product.data[CCN].meta
         # delete the generic column for conversion
         del product.data[CCN]
+        # delete the calibration key from meta as it is now processed
+        del product.data[col].meta["PCF_CURTX"]
+
         assert c == len(product.data)
```
samaloney, line 177: asserts are not guaranteed to be evaluated in need and `if else` or `try except`

PR #109, opened 2021-04-23
rework timing to always relative 

nicHoch:


samaloney:
It would be good to keep some of the tests added in this pr so keeping it open for now.

PR #117, opened 2021-06-01
add force version option to idb manager + idb version rework

nicHoch:
forcing a IDB version to be used by the manager

a version in the current IDB folder

`idb = IDBManager(test_data.idb.DIR, force_version='2.26.35')`

or any other location

`idb = IDBManager(test_data.idb.DIR, force_version=Path("..../idb.sqlite"))`


stixcore/idb/manager.py
```diff
@@ -139,14 +188,14 @@ def download_version(self, version_label, force=False,
             try:
                 Path(self._get_filename_for_version(version_label)).unlink()
             except Exception as e:
-                logger.warn(e)
+                logger.warning(e)
 
         vlabel = (IDB_VERSION_PREFIX + IDBManager.convert_version_label(version_label))
         vdir = self.data_root / vlabel
         try:
             if not vdir.exists():
```
samaloney, line 196: `vdir.mkdir( parents=True, exist_ok=True)` will create if doesn't exit and do nothing if it does

stixcore/idb/manager.py
```diff
@@ -23,22 +21,67 @@
 IDB_VERSION_DELIM = "."
 IDB_VERSION_HISTORY_FILE = "idbVersionHistory.json"
 
+IDB_FORCE_VERSION_KEY = '__FORCE_VERSION__'
```
samaloney, line 24: with a module level variable if there happen to be two instances of the class they would both access the same variable I know it's probably unlikely but any issues in making it a class variable?

stixcore/idb/manager.py
```diff
@@ -166,6 +215,61 @@ def download_version(self, version_label, force=False,
...
+        if force:
+            try:
+                Path(self._get_filename_for_version(version_label)).unlink()
+            except Exception as e:
+                logger.warning(e)
+
+        vlabel = (IDB_VERSION_PREFIX + IDBManager.convert_version_label(version_label))
+        vdir = self.data_root / vlabel
+        try:
+            if not vdir.exists():
```
samaloney, line 255: `vdir.mkdir( parents=True, exist_ok=True)`

stixcore/idb/manager.py
```diff
@@ -394,3 +503,13 @@ def get_idb(self, version_label='2.26.34', obt=None):
             return idb
         raise ValueError(f'Version "{version_label}" not found in: '
                          f'"{self._get_filename_for_version(version_label)}"')
+
+
+if __name__ == '__main__':
```
samaloney, line 508: Will this code every be executed as main?

stixcore/tmtc/packet_factory.py
```diff
@@ -115,7 +116,10 @@ class ValidationFunctionError(AttributeError):
     """
 
 
-# Main packet class
-idbm = IDBManager(Path(__file__).parent.parent / "data" / "idb")
-GenericPacket.idb_manager = idbm
+if __name__ == "__main__":
```
samaloney, line 119: Will this code every be called as main?

samaloney (changes requested):
Looks good, just a few comments. I don't think those modules can/will ever be called as `__main__` and if there were they would just execute those few lines and return?

stixcore/idb/manager.py
nicHoch: no it is just a constant name in order to lookup in a dictionary. each instance has it own dict.

stixcore/tmtc/packet_factory.py
nicHoch: separated it into test and prod code blocks

samaloney (approved):

PR #120, opened 2021-06-11
doc pipeline sync from SOC

nicHoch:


samaloney (approved):

docs/pipelineconfiguration.rst
```diff
@@ -0,0 +1,35 @@
...
+    `/opt/stixcore` :
+
+Sync with SOC
+-------------
+
+SOC TM data is synced from pub026 via rsync by cron every 10min.
+
+`crontab -e`
+
+`*/10 * * * * rsync -av stixcore@147.86.8.26:/home/solmoc/from_moc/*  /data/stix/SOC/incoming/ --exclude-from="/data/stix/SOC/processed_files.txt"`
```
samaloney, line 33: So I think we move things around so all TM that we request goes into `/home/solmox/tm` now we probably need to sync `/home/solmoc/from_moc` and `/home/solmoc/from_soc` as well but for the moment the TM is the most important.

PR #130, opened 2021-07-01
Create code to run in a pipeline operation

samaloney:
* Added code to watch a directory for new TM filess
* These tm files are then processed from TM to LB, L0 and L1
* Moved some config into a `stixconfig.ini` file and reader

### DIFF
--- a/stixcore/processing/pipeline.py
+++ b/stixcore/processing/pipeline.py
@@ -1,0 +1,72 @@
+import re
+import time
+import logging
+from pathlib import Path
+
+from watchdog.events import FileSystemEventHandler, LoggingEventHandler
+from watchdog.observers import Observer
+
+from stixcore.config.config import CONFIG
+from stixcore.io.soc.manager import SOCPacketFile
+from stixcore.processing.L0toL1 import Level1
+from stixcore.processing.LBtoL0 import Level0
+from stixcore.processing.TMTCtoLB import process_tmtc_to_levelbinary
+from stixcore.util.logging import get_logger
+
+logger = get_logger(__name__, level=logging.DEBUG)
+
+TM_REGEX = re.compile(r'.*PktTmRaw.*.xml$')
+
+
+class TMTCFileHandler(FileSystemEventHandler):
+    """
+    Handler to detect and process new TM files
+
+    As rsync is used to transfer the files to process need to take into account how rsync works.
+    Rsync works by first creating a temporary file of the same with name as the file being
+    transferred `myfile.xml` with an random extra extension `myfile.xml.NmRJ4x` it then
+    transfers the data to the temporary file and once the transfer is complete it them move/renames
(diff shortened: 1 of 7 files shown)


stixcore/processing/pipeline.py
```diff
@@ -0,0 +1,72 @@
...
+            raise TypeError('func must be a callable')
+        self.func = func
+
+    def on_moved(self, event):
+        if TM_REGEX.match(event.dest_path):
+            self.func(Path(event.dest_path))
+
+
+def process(path):
+    lb_files = process_tmtc_to_levelbinary([SOCPacketFile(path)])
```
samaloney, line 50: Not sure if the list of file returned will be unique - maybe need to create set before they are returned here

stixcore/io/fits/processors.py
```diff
@@ -491,11 +490,21 @@ def write_fits(self, product):
...
+                energies['e_low'] = elow * u.keV
+                energies['e_high'] = ehigh * u.keV
+
+                energy_enc = fits.connect._encode_mixins(energies)
+                energy_hdu = table_to_hdu(energy_enc)
+                energy_hdu.name = 'ENERGIES'
+                hdus.append(energy_hdu)
 
-            hdul = fits.HDUList([primary_hdu, control_hdu, data_hdu, idb_hdu, energy_hdu])
+            hdul = fits.HDUList(hdus)
```
nicHoch, line 508: was all this energy updates not already part of another PR

stixcore/processing/pipeline.py
```diff
@@ -0,0 +1,72 @@
...
+    l0_proc = Level0(CONFIG.get('Paths', 'tm_archive'), CONFIG.get('Paths', 'fits_archive'))
+    l0_files = l0_proc.process_fits_files(files=lb_files)
+    l1_proc = Level1(CONFIG.get('Paths', 'tm_archive'), CONFIG.get('Paths', 'fits_archive'))
+    l1_files = l1_proc.process_fits_files(files=l0_files)
+    logger.debug(l1_files)
```
nicHoch, line 55: here somewhere we might have to "update" the latest SpiceKernel

nicHoch (approved):
i think the direction is clear to me.
looking forward so see it in action

stixcore/io/fits/processors.py
samaloney: Yea it was but it assumed the energy extension was always present I just made it optional.

stixcore/processing/pipeline.py
samaloney: Yea it depends on our strategy the configuration file can also be written so could update when I new spice kernel arrives or just always use the most recent meta kernel?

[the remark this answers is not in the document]
samaloney: where should we move it to the data archive under a TM folder?

PR #132, opened 2021-07-20
Fix remaining issues process all TM to date

samaloney:


stixcore/data/idb/idbVersionHistory.json
```diff
@@ -11,14 +11,14 @@
...
         },
         {
                 "version":"2.26.35",
                 "aswVersion":181,
                 "comment": "return this for the next future",
                 "validityPeriodUTC":["2022-01-01T00:00:00", "2030-01-01T00:00:00"],
-                "validityPeriodOBT":[{"coarse":694310300, "fine":0}, {"coarse":4294967295, "fine":0}]
+                "validityPeriodOBT":[{"coarse":677766880, "fine":0}, {"coarse":4294967295, "fine":0}]
 
         }
```
nicHoch, line 23: same comment as last PR i get a bit lost about the actual times but i thought i set the "right" ones some time ago

nicHoch (approved):
is looking good

stixcore/data/idb/idbVersionHistory.json
samaloney: yea I'm not sure either but whenI was processing the TM from around the FSW update I ran into issues and then looked at the HK tm and put in rough values?

Issue #140, opened 2021-08-15
Add all IDBs for all used FSW versions used in flight 167, 174, 178, 179, 181


samaloney:
In order to process all existing TM need corresponding IDBs for FSW versions  167, 174, 178, 179, 181 and correct idbVersionHistory.json entries related to #137. 

SCET Coarse, SCET Fine, SW Version
[(640174457, 5493, 167),
 (640190310, 19341, 167),
 (640195652, 64484, 174),
 (640196996, 60980, 174),
 (640198040, 39974, 167),
 (640263339, 6606, 167),
 (640265397, 12171, 174),
 (642156300, 26757, 174),
 (642159994, 60657, 178),
 (643102337, 26882, 178),
 (643108360, 23655, 179),
 (677764274, 28479, 179),
 (677774250, 63794, 181),
 (682134257, 5399, 181)]

- [x] add versions to idbVersionHistory.json
- [x] add zip files to prod
- [x] document the process to "public" a new IDB version

samaloney:
@nicHoch I think this is close with #144 ?

PR #144, opened 2021-08-17
add older IDB versions and documentation

nicHoch:


docs/idb_versioning.rst
```diff
@@ -0,0 +1,47 @@
+IDB Versioning
+==============
+
+After a new IDB Version has been released it should be added to stixcore. All awailable IDB versions to stixcore are managed in the file `stixcore\data\idb\idbVersionHistory.json`
```
samaloney, line 4:
```suggestion
After a new IDB Version has been released it should be added to stixcore. All available IDB versions to stixcore are managed in the file `stixcore\data\idb\idbVersionHistory.json`
```

samaloney (approved):
Looks good. 

I wonder if we need to handle multiple periods for the same version? E.g 167 has to valid time ranges it was run once on the nominal side and once on the redundant side. Could be important if we want to use the same approach other things.

nicHoch:
the current code-base should support multiple time periods for the same IDB version as long as they do not overlap.

nicHoch:
fix: https://github.com/i4Ds/STIXCore/issues/140

Issue #13, opened 2020-12-02
local documentation build does not update class outline


nicHoch:
building the documentation locally: https://stixcore--9.org.readthedocs.build/en/9/developers.html#documentation

The build does not clean the class outline files in the `docs\api\` folder. Any updates to the class structure "renamed methods...." will not show up in the generated documentation.

the build/make step should include a clean

PR #155, opened 2021-09-28
Feat doc clean

samaloney:
Add a target that remove generated folders addresses  #13 not done in make as it can be really slow.


nicHoch (approved):
all good

Issue #93, opened 2021-03-26
Loop over house keeping parameters to create fits columns


samaloney:
Could probably replace this and the similar block above with a loop using a modified version of the the parameters name from the IDB? Override one that need to be modified or fix in idb

_Originally posted by @samaloney in https://github.com/i4Ds/STIXCore/pull/89#discussion_r598889779_

samaloney:
#128 does this would just need to update use the same approach for HK as well.

nicHoch:
fixed with #156

PR #156, opened 2021-09-28
simplify hk parameter list to generic loop

nicHoch:
will fix: #93


stixcore/products/tests/test_housekeeping.py
```diff
@@ -25,6 +26,12 @@
                 '0660258881:33104', '0660258881:33104', 1)]
 
 
+def setup_module():
```
samaloney, line 29: I would just use the tmp_dir or tmp_path fixture from pytest removes all this code https://docs.pytest.org/en/6.2.x/tmpdir.html

stixcore/products/tests/test_housekeeping.py
```diff
@@ -60,13 +66,12 @@ def test_calibration_hk(levelb, idbm):
     hkl0 = MaxiReportL0.from_levelb(levelb)
     hkl1 = MaxiReportL1.from_level0(hkl0)
 
-    fits_procl1 = FitsL1Processor(Path(tempfile.gettempdir()))
+    fits_procl1 = FitsL1Processor(Path(tempfile.gettempdir()) / "stixcorefits")
```
samaloney, line 69: If using tmp_path you it will generate a unit dir path for each test

samaloney (approved):

stixcore/products/tests/test_housekeeping.py
nicHoch: thx nice fixture

PR #166, opened 2021-10-22
Rework parrallelisation

samaloney:
Original version was based on totally parallel each packet was processed but this lead to issue where the same fits file was being read and written to at the same time despite efforts to lock them. This version is parallel by data type e.g. each unique combination of service type, service subtype and said are processed in parallel.

In the future really need, thread based reading with read/write queue and process based processing with a process queue. When reading check if in the write query and just return and vica versa.


nicHoch (approved):
ok this looks good
just minor comment

Issue #170, opened 2021-11-12
Dropped empty packets in QL cause crash when processing from LB to L0 


samaloney:
In quicklookL0.py we have this code 

```
packets, idb_versions = GenericProduct.getLeveL0Packets(levelb)

control = Control.from_packets(packets, NIX00405_offset=NIX00405_offset)
control['raw_file'] = levelb.control['raw_file']
control['packet'] = levelb.control['packet']
control['parent'] = parent
```

but number of packets after parsing can be less than the number of input packet, due to empty packets being dropped, so the length of `control['raw_file']` and `levelb.control['raw_file']` can be of different lengths, same for `control['packet']` an`levelb.control['packet']`. Need to match up using some common data

samaloney:
Closed in #171

Issue #84, opened 2021-03-11
Aspect implement fix for FSW for older packets 


samaloney:
See https://github.com/i4Ds/STIX-FSW/issues/934 basically need to handle 3 cases:

before NIXxxx was added assume value 
after NIXxx was added but before FSW fix
after NIXxx was addd and FSW fix

PR #171, opened 2021-11-12
FITs improvements and processing bug fixes

samaloney:
Main goal of this PR is to ensure the generate fits file work well with IDL mrdfits (if using unsigned int trick, make sure bscale is set to 1). 

Also reduced the fits file sizes where possible by using a a dynamic, per fits file, uint size using the max values of the for the counts and triggers. For example often for CPD bytes or uint16 are sufficient but for spectrogram triggers sometime uint64 are needed.

Finally addressed some small bugs found while testing on all current TM


stixcore/ephemeris/manager.py
```diff
@@ -158,8 +158,7 @@ def __init__(self, meta_kernel_path):
 
             self.mod_meta_kernel_path = Path(path)
 
-        # *_, datestamp, version = self.meta_kernel_path.name.split('_')
-        # self.kernel_date = datetime.strptime(datestamp, '%Y%m%d')
+        self.file = self.meta_kernel_path.name
```
samaloney, line 161: @nicHoch if we make sure the spice manger  always points to the dated version of the MK this should work at the moment it work but always points to `solo_ANC_soc-flown-mk.tm`  but if we use the other file e.g. `solo_ANC_soc-flown-mk_v107_20210909_001.tm` I think this will work.

samaloney:
Closes #170, #161 and partially addressed #84

nicHoch (approved):
i thought id did already approve

PR #176, opened 2021-11-19
Bugfix fits l1 dtypes

### LINKED ISSUE #174: L1 fits files don't maintain the same types as L0 all int uint of any size become float64
There is a bug with in astropy where when QTable are read in [see this issue](https://github.com/astropy/astropy/issues/12494) so have to add some kind of work around as the dramatically increase science data file size by about ~10

samaloney:
Closes #174 by implement a custom QTable reader until/if upstream fix happens.


nicHoch (approved):
do i have to use the new table read method also in the dpdd generator - so that the right type is used?

samaloney:
I'm not sure, depends on how you are reading the fits files and pull out the information?

PR #177, opened 2021-11-22
Spice manager global

nicHoch:
- removed time and position spice context
- combined into a single "Spice" class
- not a context anymore as this has to be propagated to the sub processes for parallel computing 
- the global singleton will be auto instantiated for testing but should be set up manually for any __main__


stixcore/processing/L0toL1.py
```diff
@@ -74,6 +81,10 @@ def process_type(files, processor):
     fits_path = Path('/home/shane/fits_test/L0')
     bd = Path('/home/shane/fits_test/')
 
+    # possible set an alternative spice kernel if not the latest should be used
+    SpiceKernelManager.instance = SpiceKernelManager(Path(CONFIG.get("Paths", "spice_kernels")))
```
samaloney, line 85: Does this need to be a singleton?

samaloney (approved):

samaloney:
Also just wonder how this will work with IDL will you give it the path or set it when calling from hissw?

stixcore/processing/L0toL1.py
nicHoch: hmm right the manager do not really have to be a singleton.

just the Spice object need to be one if we will avoid the pass it around everywhere 

will fix this

nicHoch:
> Also just wonder how this will work with IDL will you give it the path or set it when calling from hissw?

We can decide if we want to pass the path into hissw or use the environment variable. both ways are ready to use.

PR #181, opened 2021-11-23
use raw data for L0 in EnergyCalibration

nicHoch:


samaloney (approved):

samaloney:
Does this then get converted in L0toL1 as expected?

nicHoch:
> Does this then get converted in L0toL1 as expected?

uiiii good point.
i was assuming so. but checked anyway and surprise surprise at the moment only columns in data where out converted not in control ;-(

PR #182, opened 2021-11-23
Fix bug in calibration spectra processing LB to L0

### LINKED ISSUE #168: Energy Calibration Data Error when pixel mask changes in a day
`l0_files = l0processor.process_fits_files(files=[Path('/home/shane/fits_test_local/LB/21/6/41/solo_LB_stix-21-6-41_0642124800_V01.fits')])`

```python
np.unique(control['pixel_mask'], axis=0)

array([[1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]], dtype=uint8)
```

Essentially need to treat each complete packet separately as when the pixel mask/or detector mask changes the current code breaks down prob need to treak more like pixel data and loop of the create an full array (32, 12, 1024) and loop over the detector and pixel ids as given in the tm

samaloney:
Closes #168

### DIFF
--- a/stixcore/processing/LBtoL0.py
+++ b/stixcore/processing/LBtoL0.py
@@ -133,8 +133,8 @@ if __name__ == '__main__':
 
     warnings.filterwarnings('ignore', module='astropy.io.fits.card')
 
-    fits_path = Path('/home/shane/fits_test_local/LB/21/6/42')
-    bd = Path('/home/shane/fits_test_local')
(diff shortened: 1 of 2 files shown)


nicHoch (approved):
hmmm hard to follow up here
not sure if i get all dimensions right but code wise looks good to me

samaloney:
Yea I processed all the calibration TM with this branch and at least got no errors but should probably check some.

PR #183, opened 2021-11-24
Fix #115 update detector mask for affect times

### LINKED ISSUE #115: Masks in BSD X-ray date are not representative for certain time periods
Due to the configuration of the SumDMask parameter the CFL and BK detector data was not being recorded to the archive buffer so even if data is requested it will contain zeros. For normalisation purposes it would be better to adjust the detector mask on ground to reflect this. Date for this are late November 2020 to mid March 2021 @grazwegian

1.  - [x]  Get exact date for the problem - possibly query tc for tm or parameter change?
2.  - [x] Add case to override the detector masks
3. - [ ] Add test

samaloney:
Closes #115


samaloney:
Closes #115

nicHoch (approved):
nice and clean also with documentation 
👍

PR #184, opened 2021-11-25
add the control table to the raw-eng processing

nicHoch:


samaloney (approved):
Looks good - I assume the L1 files should be unchanged by this this?

PR #185, opened 2021-11-25
Final fits file size reduction efforts

### LINKED ISSUE #141: Fits file size review
Overview of sizes of various fits archives

---

### TM to LB
| level | size | status | notes |
|------|-----|-------|-------|
| TM | 3.6G |   |  |
| LB | 1.5G | 🟡 | most likely due to removal of duplicate tm but need to verify |

---

### LB to L0/L1

| service | LB | L0 | status | notes |
|--------|----|----|-------|-------|
| 1 | 9.4M | 7.1M | 🟠  | Missing some TM due to IBD |
| 3 | 198M | 381M | 🟡 | Probably ok not sure where size increase comes from |
| 5 | 7.1M | 8.1M | 🟢 |  |
| 6 | 88K | 104K | 🟡 | Probably ok not sure where size increase comes from | 
| 17 | 188K | 188K | 🟢 | |
| 21 | 1.3GB | 14GB | 🟠 | see break down below |
| 237 | 28K | 32K | 🟡 | Probably ok not sure where size increase comes from |
| 238 | 92K | 128K | 🟡 | Probably ok not sure where size increase comes from |
| 239 | 1.1M | 5.9M | 🟠 | seems a lot | 

21,6 by SSID

| SSID | LB | L0 | status | notes |
|------|----|---|--------|-------|
| 30 | 88M | 716M | 🟠  | seems a lot should be mainly decompression check data type in fits files |
| 31 | 39M | 78M | 🟡   |  |
| 32 | 58M | 85M | 🟡  |  |
| 33 | 19M | 61M |  | 🟡 |  |
| 34 | 37M | 48M | 🟡  |  |
| 41 | 157M | 2.6G | 🔴   |  investigate no compression so would have expected closer match |
| 20 | 3.5M | 66M |  🔴  | investigate  |
| 21 | 733M | 9.2G | 🔴 | investigate  |
| 22 | 260K | 716K |  🟡  |   |
| 23 | 220K | 384K | 🟡  |   |
| 24 | 39M | 334M | 🔴  |   |
| 42 | 18M | 30M | 🟠  | aspect isn't compressed to should be very similar - could be overhead  |

Need to check on data types in fits files make as small as possible, in particular I think `astropy.quantites` default to 64bit floats. Expect some overhead due to fits headers etc but some be dominated by decompression need to figure out what we expect and then compare to measured.

For example in investigating #136 I discovered both `time` and `timedel` are being stored as 64bit floats only need to be 32bit uints or similar 

Possible improvements:
- time uint32 (really on 24 bits) and timedel (probably less)
- for BDS counts at during io check max value and cast to min possible dytp e.g uint16, uint32, or unint64
- Leave text and other parameter as bits and use Enums

samaloney:
* Fix a few remaining places that defaulted to int64 or float64
* in RPD and CPD often ~50% of the entires were zeros due to alway reshaping  to t x32 x 12 x 32, now only store up to max energy as often have significantly less than 32 so now each file will vary t x 32 x 12 x n where n is the max energy bin requested
* Also change some log level of specific message and of the loggers

Should close #141


samaloney:
See https://github.com/i4Ds/STIXCore/issues/141#issuecomment-979282931

stixcore/products/product.py
```diff
@@ -169,7 +169,7 @@ def __call__(self, *args, **kwargs):
                     try:
                         energies = QTable.read(file_path, hdu='ENERGIES')
                     except KeyError:
```
nicHoch, line 171: should here also the new read table wrapper function be used?

nicHoch (approved):
looks good. but mainly i have to trust the tests

stixcore/products/product.py
samaloney: Possibly I don't think we ever use the energies which are read we always recreate when writing but no harm to use it.

samaloney: ✅

PR #189, opened 2021-11-30
Bugfix Spectrogram RCR and some QL units

### LINKED ISSUE #188: Spectrogram files only contain one RCR value per file
Spectrogram TM doesn't contain an RCR values per time bin so need to generate on the ground. Ideally would like to do using spectrogram file contents only but this probably won't be sufficient. Could look at HK data to find RCR state transition and then user a derivative of the count to find exact time bin in spectrogram data or if L1 data is present at max cadence could use this alternatively.

samaloney:
Add rcr to data table similar to other BSD fixes #188


nicHoch (approved):
nice that u.ct exists 
from a science point of view is that really a unit?

samaloney:
No I guess it's not physical unit but guess there were enough use cases for it to be added.

PR #195, opened 2021-12-15
Follow up to IDB36 update

samaloney:
* Fix idb folder structure issue
* Update idb version history with real dates
* Fix tiny when round time for fits file in L1 serialisation


nicHoch (approved):
looks good i am not sure why this now in a PR of you as most of the changes came from me.

Issue #135, opened 2021-08-04
Energy units could already be added the get_energies() method


samaloney:
could the units not already get attached in the get_energies() method?

_Originally posted by @nicHoch in https://github.com/i4Ds/STIXCore/pull/128#discussion_r674679324_

PR #201, opened 2022-01-18
encapsulate  energy extension creation

nicHoch:
fix: https://github.com/i4Ds/STIXCore/issues/135 and #202


samaloney (changes requested):
Could you add the default tunit to the energy extension to this as well? similar to this `data_hdu = add_default_tuint(data_hdu)` was some feedback from ESA on latest fits files

PR #203, opened 2022-02-04
Plattform integration

nicHoch:
add function to define a common interface to the "Platform"

### DIFF
--- a/stixcore/tmtc/parameter.py
+++ b/stixcore/tmtc/parameter.py
@@ -1,6 +1,10 @@
 from collections import defaultdict
 
+import numpy as np
+
 import astropy.units as u
+
+from stixcore.time.datetime import SCETime
 
 __all__ = ['Parameter', 'EngineeringParameter', 'CompressedParameter']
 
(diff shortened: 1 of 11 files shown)


stixcore/tmtc/parameter.py
```diff
@@ -17,15 +21,68 @@ class Parameter:
...
+            elif self.name == 'NIX00402':
+                if self.value > 2 ** 32 - 1:
+                    coarse = self.value >> 16
+                    fine = self.value & (1 << 16) - 1
+                else:
+                    coarse = self.value
+                    fine = 0
+                val = SCETime(coarse, fine).to_datetime().isoformat(timespec='milliseconds')
+            elif self.name == 'NIXD0003':
+                val = round(self.value / 2.5, 3)
```
samaloney, line 61: Is this to match Hualin's code or do we have something wrong?

stixcore/tmtc/tests/test_packets.py
```diff
@@ -61,6 +62,13 @@ def test_tc_packet():
                 for key in TC_DATA_HEADER_STRUCTURE.keys() if not key.startswith('spare')])
 
 
+def test_packet_laszlo(idb):
```
samaloney, line 65: maybe a better name or description in comments?

stixcore/products/level0/quicklookL0.py
```diff
@@ -173,7 +174,8 @@ def from_levelb(cls, levelb, parent=''):
                    ssid=packets.ssid,
                    control=control,
                    data=data,
-                   idb_versions=idb_versions)
+                   idb_versions=idb_versions,
+                   packets=packets)
```
samaloney, line 178: So is this so you can directly pass in a list of packets rather than a `Product`?

samaloney (approved):

samaloney:
Maybe delete the workflow files from this as they will be added in #213

stixcore/tmtc/parameter.py
nicHoch: this is to match Hualin's format

stixcore/tmtc/tests/test_packets.py
nicHoch: removed test

PR #213, opened 2022-03-03
End2endtest

nicHoch:
still many hard coded pathes and some errors ...
but needs to be merged so the server can produce the "original data"


samaloney (approved):
So you checkout master on the pub099 server run the script to generate the zip files and run the test. Then on the GitHub you download these file and run the test with the current code and run the comparison test?

.github/workflows/end2end_build.yml
```diff
@@ -0,0 +1,28 @@
+# This is a basic workflow to help you get started with Actions
```
samaloney, line 1: Maybe update to match what this does?

setup.cfg
```diff
@@ -22,6 +22,7 @@ install_requires =
     watchdog
     pyyaml
     intervaltree
+    GitPython
```
samaloney, line 25: Not for this PR but we should probably try split into a a few section of requirements

stixcore/ephemeris/manager.py
```diff
@@ -141,6 +141,7 @@ def __init__(self, meta_kernel_path):
 
         curdir = os.getcwd()
         try:
+            print(f"LOAD NEW META KERNEL: {self.meta_kernel_path}")
```
samaloney, line 144: logger?

stixcore/processing/tests/test_end2end.py
```diff
@@ -0,0 +1,75 @@
...
+    return list(orig_data.rglob("*.fits"))
+
+
+@pytest.fixture(scope="session")
+def current_fits(orig_data, out_dir):
+    return en2end_pipeline(orig_data, out_dir)
+
+
+@pytest.mark.end2end
+def test_compleet(orig_fits, current_fits):
```
samaloney, line 45:
```suggestion
def test_complete(orig_fits, current_fits):
```

samaloney, line 45: what does complete mean the packets of all the files are there or something else?

stixcore/processing/tests/test_end2end.py
```diff
@@ -0,0 +1,75 @@
...
+            next(cfits for cfits in current_fits if ofits.name == cfits.name)
+        except StopIteration:
+            error = True
+            warnings.warn(f"no corresponding file found for {ofits} in the current fits files")
+    if error:
+        raise ValueError("one or many errors\nnumber of fits files differ")
+
+
+@pytest.mark.end2end
+def test_identical(orig_fits, current_fits):
```
samaloney, line 58: could you have identical but incomplete ? I guess I just wondering if the previous test redundant

stixcore/util/scripts/end2end_testing.py
```diff
@@ -0,0 +1,150 @@
...
+from stixcore.data.test import test_data
+from stixcore.ephemeris.manager import Spice, SpiceKernelManager
+from stixcore.io.soc.manager import SOCManager
+from stixcore.processing.L0toL1 import Level1
+from stixcore.processing.LBtoL0 import Level0
+from stixcore.processing.TMTCtoLB import process_tmtc_to_levelbinary
+from stixcore.products.product import Product
+from stixcore.tmtc.packets import TMTC
+
+pytest.skip
```
samaloney, line 18: Is this needed on the CI?

stixcore/util/scripts/end2end_testing.py
```diff
@@ -0,0 +1,150 @@
...
+    newroot = [Et.Element('Response') for i in splitfiles]
+    newresponse = [Et.SubElement(newroot[i], 'PktRawResponse') for i in splitfiles]
+
+    for f, u in files:
+        try:
+            l1_f = Path(f)
+            l1_p = Product(l1_f)
+            rootdir = l1_f.parent.parent.parent.parent.parent.parent
+
+            l0_f = (rootdir / "L0").rglob(l1_p.parent[0]).__next__()
```
samaloney, line 76: I don't think there is any guarantee of wording form `glob` so it may very from OS to OS of because  of access or modify date

stixcore/util/scripts/end2end_testing.py
```diff
@@ -0,0 +1,150 @@
...
+                rpa.set("packetID", str(packetid))
+                pa = Et.SubElement(rpa, "packet")
+                pa.text = list(node)[0].text
+        except Exception as e:
+            print(e)
+    print(f"{packetid} packets recompiled")
+
+    for i, root in enumerate(newroot):
+        with open(outdir / f'Test{i}.PktTmRaw.xml', 'wb') as xmlf:
+            _pretty_xml_print(root)
```
samaloney, line 104: lxml has a pretty print function but maybe it not what you need `etree.tostring(x, pretty_print=True)`

stixcore/util/scripts/end2end_testing.py
```diff
@@ -0,0 +1,150 @@
...
+            tree.write(xmlf, encoding='utf-8')
+
+    fitsdir = outdir / "fits"
+    if fitsdir.exists():
+        shutil.rmtree(str(fitsdir))
+    print(f"saved: xml files {len(newroot)}")
+
+    fitsfiles = en2end_pipeline(outdir, fitsdir)
+    for f in fitsfiles:
+        print(f)
```
samaloney, line 115: Is there a reason to use `print` over `logging` or just for convenance?

stixcore/util/scripts/rebuild_hook.cgi
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/bash
```
samaloney, line 1: Yea I think what you probably needed was the cgi module e.g. (https://stackoverflow.com/questions/44871139/how-do-i-run-python-cgi-script-on-apache2-server-on-ubuntu-16-04) to connect the python to the web server but this works too.

stixcore/util/scripts/rebuild_hook.cgi
```diff
@@ -0,0 +1,61 @@
...
+
+echo "Recreate the test files at $ZIPPATH and pack them afterwards to: $DATAPATH"
+
+CMD="$DIR/venv/bin/python"
+SCRIPT="$DIR/stixcore/util/scripts/end2end_testing.py"
+
+echo "<br />cmd# $CMD $SCRIPT $ZIPPATH $DATAPATH<br />"
+nohup $CMD $SCRIPT $ZIPPATH $DATAPATH &>/dev/null &
+
+echo "<h2>all Done zip will be awailable soon</h2>"
```
samaloney, line 61:
```suggestion
echo "<h2>Complete, zip will be available soon</h2>"
```

samaloney (approved):

setup.cfg
nicHoch: removed as the cgi hook is not a python file any more

.github/workflows/end2end_build.yml
nicHoch: updated the comments

stixcore/util/scripts/end2end_testing.py
nicHoch: removed

nicHoch: this is needed to keep a reference to pytest "alive" as isort/flake8 will remove it otherwise
but this script assumes that the test config to be loaded
i added a comment

stixcore/util/scripts/end2end_testing.py
nicHoch: print to stdout is used as the output of this script might get redirected to the webhook

PR #217, opened 2022-03-07
Fits updates from ESA SOAR and DPDD review

samaloney:


nicHoch (approved):
this is looking straight forward

samaloney:
I just processed all current TM with this code and did see any additional problems and changes in fits files looked good.

PR #220, opened 2022-03-21
Plattform integration test

nicHoch:
somehow some tests got lost 
readd them


samaloney (approved):
Looks goods not sure why all those extra commit are still showing up.

PR #221, opened 2022-03-22
ddpd

nicHoch:
semi automatic generation of the ddpd document


nicHoch:
i think the rebase before the PR is not a good way to go as it looks like this PR has changed many files (also the end2end fails) but this the main files changed are:

stixcore/data/test/ddpd.in.csv  # descriptions override text file

stixcore/products/__init__.py # add low latency products to Factory

stixcore/products/lowlatency/quicklookLL.py # provide products for low latency 

stixcore/products/product.py # provide products for low latency 

stixcore/util/scripts/ddpd.py # the actual dpdd code

samaloney (approved):

PR #226, opened 2022-04-11
Fix bug in QL Background extraction

samaloney:
(cherry picked from commit f12238de507841d858c5d492a2c417c967180229)


samaloney:
No sure how this happened again but old code got in there and was not processing the bkg data properly

nicHoch (approved):

PR #232, opened 2022-05-20
Bsd in mult tm files

nicHoch:
see #169 

if TM xml files are not processed in order or if the TM data from the same day split up into multiple part (as the size is limited to ~50MB) some BSD TM products where not write out to FITS or are "shorter". 

The new detect for complete packet sequences logic should fix that. the LB packets are not processed any further until all packets are available later on. so this LB fits file will be updates later on (other day) from the pipeline. and the processing is re triggered. all already generated L0... data fits file will not be combined/overridden but the so fare not processed BSD packet sequence ( now compete) will we written to fits.

Right now the complete test checks if the number of packets in the sequence start, n ... m, end is the number of the sequence_count[end] - sequence_count[start].

I did a FSW review. following scenario is possible.

a BSD request is processed **start** was send. Now the mode changes to save no **end** is send. Later on the request is processed again (continued). A new **start** is send and in the end also the **end** signal.

so a sequence will look like: start(sequence_count=1), 2, 3, 4, start(sequence_count = ?), ?+1, ?+2, end(sequence_count+3).

the sequence count will only be continuous if the restarted BSD request processing is the first in the row. This somehow should be the case looking into the code. BUT it is not guaranteed.


stixcore/products/level0/scienceL0.py
```diff
@@ -99,24 +99,26 @@ def parent(self):
...
-        # return type(self)(service_type=self.service_type, service_subtype=self.service_subtype,
-        #                   ssid=self.ssid, data=data, control=control)
-        raise(ValueError(f"Tried to combine 2 BSD products: {self} and {other}"))
+        if self.complete or other.complete:
+            raise(ValueError(f"Tried to combine 2 BSD products: \n{self} and \n{other}"))
+
+        if (np.all(self.control == other.control) and self.scet_timerange == other.scet_timerange
+                and len(self.data) == len(other.data)):
+            return self
+        combined_control_index = other.control['index'] + self.control['index'].max() + 1
```
samaloney, line 108: This seems to assume other and self are ordered what if `other.control['index'] = [4, 5, 6]` and `self['index'] = [1, 2, 3]` for example?

stixcore/products/level0/scienceL0.py
```diff
@@ -99,24 +99,26 @@ def parent(self):
...
+        cnames = control.colnames
+        cnames.remove('index')
+        control = unique(control, cnames)
+
+        combined_data_index = other.data['control_index'] + self.control['index'].max() + 1
+        data = vstack((self.data, other.data))
+        # TO DO Go on here
+        data_ind = np.isin(combined_data_index, combined_control_index)
+        data = data[data_ind]
+
```
samaloney, line 119: I think the control and data indices need to be rebuilt here so `control['index`]` is monotonic from 0 to n then replace the `data[control_index]` with the corresponding new index

samaloney (changes requested):
Look goods I think there might be some edge cases in the add method that need to be handled but not sure.

Also I don't understand where all these commit are coming from there should only be on three?

stixcore/products/level0/scienceL0.py
nicHoch: is removed

stixcore/products/level0/scienceL0.py
nicHoch: also removed

PR #238, opened 2022-06-09
Pipeline logging

nicHoch:


stixcore/processing/pipeline.py
```diff
@@ -50,36 +58,163 @@ def __init__(self, func, regex):
         if not isinstance(regex, type(re.compile('.'))):
             raise TypeError('regex must be a regex Pattern')
         self.regex = regex
+        self.args = args
+        self.queue = Queue(maxsize=0)
+        self.name = name
+        self.lp = threading.Thread(target=self.process)
+        self.lp.start()
+
+    @poll_decorator(step=1, poll_forever=True)
```
samaloney, line 67: So you poll but only will the `queue.get` not wait forever at each poll?

nicHoch: yes the  queue.get will wait until something new comes in.
But after that the process is done.
the poll_decorator starts a new process afterwards again and again .

stixcore/processing/pipeline.py
```diff
@@ -53,35 +59,145 @@ def __init__(self, func, regex, **args):
...
+        logging.getLogger().removeHandler(self.fh)
+        if not self.allright:
+            shutil.copyfile(self.log_file, self.err_file)
+            if CONFIG.getboolean('Pipeline', 'error_mail_send', fallback=False):
+                try:
+                    sender = CONFIG.get('Pipeline', 'error_mail_sender', fallback='localhost')
+                    receivers = CONFIG.get('Pipeline', 'error_mail_receivers').split(",")
+                    host = CONFIG.get('Pipeline', 'error_mail_smpt_host', fallback='localhost')
+                    port = CONFIG.getint('Pipeline', 'error_mail_smpt_port', fallback=25)
+                    smtpObj = smtplib.SMTP(host=host, port=port)
```
samaloney, line 139:
```suggestion
                    smtp_server = smtplib.SMTP(host=host, port=port)
```

stixcore/processing/pipeline.py
```diff
@@ -53,35 +59,145 @@ def __init__(self, func, regex, **args):
...
+
+{self.err_file}
+
+StixCore
+
+==========================
+
+do not answer to this mail.
+"""
+                    smtpObj.sendmail(sender, receivers, message)
```
samaloney, line 154: For test you could mock this method and just make it has the right contents. Now if you really want to test if the email is sent that won't work but I think that would be an external issue to stixcore

stixcore/processing/tests/test_processing.py
```diff
@@ -268,11 +270,69 @@ def test_single_vs_batch(out_dir):
...
+    finally:
+        CONFIG.set('Logging', 'stop_on_error', str(CONTINUE_ON_ERROR))
+        CONFIG.set('Paths', 'fits_archive', str(FITS_ARCHIVE))
+        CONFIG.set('Pipeline', 'log_level', str(LOG_LEVEL))
+        CONFIG.set('Pipeline', 'log_dir', str(LOG_DIR))
 
-    currently used to start the AUX processing on an existing L1 fits dir
-    '''
+
+@pytest.mark.skipif(not CONFIG.getboolean('Pipeline', 'error_mail_send', fallback=False),
```
samaloney, line 304: As mentioned above you could mock this so the test could run

tox.ini
```diff
@@ -42,7 +42,7 @@ deps =
 extras =
     test
 commands =
-    pytest --pyargs stixcore {toxinidir}/docs --cov stixcore --cov-config={toxinidir}/setup.cfg --remote-data=any -m "not end2end" {posargs}
+    pytest --pyargs stixcore {toxinidir}/docs --cov stixcore --cov-config={toxinidir}/setup.cfg --remote-data=any -m "not end2end" -n 1 --dist loadfile {posargs}
```
samaloney, line 45: Is the distributed test part of this PR?

stixcore/idb/tests/test_idb.py
```diff
@@ -17,10 +17,12 @@
 
 
 @pytest.fixture
+@pytest.mark.remote_data
```
samaloney, line 20: maybe in the test config we should download the IDB so this isn't necessary? seem awkward

setup.cfg
```diff
@@ -23,6 +23,7 @@ install_requires =
     pyyaml
     intervaltree
     hissw
+    polling2
```
samaloney, line 26: I guess I'm a little bit concerned about having a dependancy on a one man library

stixcore/processing/pipeline.py
samaloney: What the advantage compare to doing something like this inside the function? 
```
while True:
    path = self.queue.get()
    try:
         self.func(path, **self.args)
    except Exception as e:
        logger.error(e)
             if CONFIG.getboolean('Logging', 'stop_on_error', fallback=False):
                 raise e

samaloney (approved):

tox.ini
nicHoch: the preparations somehow yes 
but n is set to 1 so ineffective at the moment

setup.cfg
nicHoch: yes but this is a very isolated function and location in the code
if problems show up we can replace it

stixcore/processing/tests/test_processing.py
nicHoch: good point will do so

PR #240, opened 2022-06-16
Update transmission components and fix bug

samaloney:
* Update model with new information on alloy composition for attenuator and windows
* Fix small bug in tranmission code


nicHoch (approved):
quite a interesting model

samaloney (commented):
Looks good to me

PR #259, opened 2022-08-10
change parallel processing to batch by SCI product

nicHoch:
a new parallel processing grouping: 

Groups all files into parallel processes by product type if daily products. For by request products it is also grouped by type but additional into batches of a given size.
In L1 processing: also a simple priority heuristic starts with ql data as this takes longest.

takes down: 
L0 processing from 48h to 9h
L1 processing from 55min to 42min
L2 processing already implemented but needs measurements and maybe adjustments


samaloney (commented):
Looks ok, once the tests pass

stixcore/util/logging.py
```diff
@@ -20,10 +20,14 @@
...
 """
 import logging
+import warnings
 
 STX_LOGGER_FORMAT = '%(asctime)s %(levelname)s %(name)s %(lineno)s: %(message)s'
 STX_LOGGER_DATE_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
 
+warnings.filterwarnings('ignore', module='astropy.io.fits.card')
+warnings.filterwarnings('ignore', module='astropy.utils.metadata')
+
```
samaloney, line 30: I'm not sure this is the best place for this as this will turn off in all modules prob only should do in the pipeline?

stixcore/util/scripts/daily_comparison.py
```diff
@@ -12,6 +12,31 @@
...
+        while parent != pathL0:
+            subd.insert(0, parent.name)
+            parent = parent.parent
+        for s in subd:
+            target = target / s
+        found = len(list(target.rglob(name)))
+        if found == 0:
+            print(f"{f} >> {target} .. {name}")
+
+if __name__ == '__main__2':
```
samaloney, line 39: never seen `'__main__2'`

samaloney (approved):
Once the test pass should be good to go.

stixcore/util/scripts/daily_comparison.py
nicHoch: this is just an alternate main that i rename from time to time will clean it up

stixcore/util/logging.py
nicHoch: moved

PR #262, opened 2022-08-16
shring the end2end test files the a minimum

nicHoch:
after the fix of https://github.com/i4Ds/STIXCore/pull/261 i was able to back trace to the real packets and minimice the end2end test input. so the test should be much faster and more easy to track down any differences.

will fore sure not pass the current end2end


samaloney (approved):
I would always favour logging.debug, logging.info over print statements but not a show stopper

PR #270, opened 2022-08-25
End2end fix aka order independend TM processing

nicHoch:
a lot of the efforts in this PR are about to make the TM processing order independent 
mainly the product.add() method was adjusted to generate the symmetric output

existing + current == current + existing


stixcore/products/product.py
```diff
@@ -480,40 +481,85 @@ def __add__(self, other):
...
+
+        # update the index in control as used in data
+        for idx, row in enumerate(control):
+            oid = row['old_index']
+
+            if oid not in newids:
+                del_rows.append(idx)
+                continue
+            nid = newids[oid]
+            row['index'] = nid
```
samaloney, line 538: So I think this can be by just dropping the id not in the data table but we can tackle this later

samaloney (approved):

samaloney:
we should  prob make an issue
```
Headers contain differences:
[63](https://github.com/i4Ds/STIXCore/runs/8031966729?check_suite_focus=true#step:6:64)
     Keyword COMMENT [146] has different values:
[64](https://github.com/i4Ds/STIXCore/runs/8031966729?check_suite_focus=true#step:6:65)
        a>   unit: '0.01'
[65](https://github.com/i4Ds/STIXCore/runs/8031966729?check_suite_focus=true#step:6:66)
         ?          ^^^^
[66](https://github.com/i4Ds/STIXCore/runs/8031966729?check_suite_focus=true#step:6:67)
        b>   unit: '%'
[67](https://github.com/i4Ds/STIXCore/runs/8031966729?check_suite_focus=true#step:6:68)
         ?          ^
```

PR #275, opened 2022-09-07
Save fits files in folder for the day they start not mid point. 

samaloney:
So at the moment the file mid point is use to determine where it is stored this PR change this to use the start date.


nicHoch (approved):
looks good

Issue #253, opened 2022-07-12
 TM Processing L0: 21,6,24 pixel mask not unique


nicHoch:
see also: #248 

```
/data/stix/out/test/bsdfull/LB/21/6/24/solo_LB_stix-21-6-24_0000000000-9999999999_V01_2205207183-51824.fits for 21, 6, 24
/data/stix/out/test/bsdfull/LB/21/6/24/solo_LB_stix-21-6-24_0000000000-9999999999_V01_2203306924-51836.fits for 21, 6, 24

```

2022-07-09 00:11:43,507 VerifyWarning: Card is too long, comment will be truncated.
2022-07-09 00:11:46,042 Error processing file /data/stix/out/test/bsdfull/LB/21/6/24/solo_LB_stix-21-6-24_0000000000-9999999999_V01_2203306924-51836.fits for 21, 6, 24
Traceback (most recent call last):
  File "/home/nicky/STIXCore/stixcore/processing/LBtoL0.py", line 100, in process_tm_type
    level0 = tmp.from_levelb(comp, parent=file.name)
  File "/home/nicky/STIXCore/stixcore/products/level0/scienceL0.py", line 600, in from_levelb
    control['pixel_masks'] = np.unique(_get_pixel_mask(packets)[0], axis=0)
  File "/home/nicky/STIXCore/venv/lib/python3.8/site-packages/astropy/table/table.py", line 1904, in __setitem__
    self.add_column(value, name=item, copy=True)
  File "/home/nicky/STIXCore/venv/lib/python3.8/site-packages/astropy/table/table.py", line 2193, in add_column
    raise ValueError('Inconsistent data column lengths')
ValueError: Inconsistent data column lengths
2022-07-09 00:11:46,043 Inconsistent data column lengths
2022-07-09 00:12:02,695 VerifyWarning: Card is too long, comment will be truncated.

PR #277, opened 2022-09-22
Fix bug introduced in summed pixel data due to varible pixel masks

samaloney:
Fix bug introduced by fix for #253


nicHoch (approved):
looking good

PR #280, opened 2022-10-03
publish to ESA as a independend processing step

nicHoch:
add a persistent file (sqlite) where we keep track of what has been delivered to ESA so fare.

with the storage in place we can check for certain conflicts and come up with rules how to resolve.


stixcore/processing/publish.py
```diff
@@ -0,0 +1,300 @@
...
+*********
+{pformat(files[PublishConflicts.ADDED])}
+
+
+StixCore
+
+==========================
+
+do not answer to this mail.
+"""
```
samaloney, line 171: move to var defined at top of file may `PUBLISH_EMAIL_TEMPLATE = f""" ...."""`

stixcore/processing/publish.py
```diff
@@ -0,0 +1,300 @@
...
+        except Exception as e:
+            published['failed'].append((p, e))
+            logger.error(e, exc_info=True)
+
+    send_mail_report(published)
+    hist.close()
+
+
+if __name__ == '__main__':
+    publish_fits_to_esa(sys.argv[1:])
```
samaloney, line 300: So would the plan to be to run this a cron job or similar?

samaloney (approved):

stixcore/processing/publish.py
nicHoch: yes

stixcore/processing/publish.py
nicHoch: good point

stixcore/processing/publish.py
```diff
@@ -0,0 +1,671 @@
...
+
+    Parameters
+    ----------
+    args : _type_
+        _description_
+
+    Returns
+    -------
+    _type_
+        _description_
```
samaloney, line 329: looks like auto generated txt?

stixcore/processing/tests/test_publish.py
```diff
@@ -0,0 +1,180 @@
...
+                               '--db_file', str(out_dir / "test.sqlite"),
+                               '--fits_dir', str(fits_dir)])
+
+    assert res
+    # the first one was added
+    assert len(res[PublishResult.PUBLISHED]) == 1
+    # 2 where published as supplement
+    assert len(res[PublishResult.MODIFIED]) == 2
+    # 3 where ignored as the data is the same
+    assert len(res[PublishResult.IGNORED]) == 2
```
samaloney, line 177: comment doesn't match the code not sure if just not updated really important?

docs/pipelineconfiguration.rst
```diff
@@ -133,3 +133,44 @@ You might toggle the parameter only for manually restarting the service after yo
...
+                        with already published files
+  -d DB_FILE, --db_file DB_FILE
+                        Path to the history publishing database
+  -w WAITING_PERIOD, --waiting_period WAITING_PERIOD
+                        how long to wait after last file modification before publishing
+  -v INCLUDE_VERSIONS, --include_versions INCLUDE_VERSIONS
+                        what versions should be published
+  -l INCLUDE_LEVELS, --include_levels INCLUDE_LEVELS
+                        what levels should be published
+  -p INCLUDE_PRODUCTS, --include_products INCLUDE_PRODUCTS
```
samaloney, line 173: would be nice to show the examples of valid values? Is this a copy past of the running the command with `-h` if so maybe add to the actual argparse config?

samaloney (approved):
Wow this does a lot. I think the code could do with being refactored to:
1.  be a bit more pythonic 
2. improve separation of concerns publish.py does everything updating, copying, history etc

but it's working and we need it so I think this could be done at a later stage if there ever time.

docs/pipelineconfiguration.rst
nicHoch: will fix

stixcore/processing/publish.py
nicHoch: fixed

Issue #291, opened 2022-11-17
Mismatch between detector mask and shape of counts array in cpd files


FredSchuller:
In some L1 pixel data files (solo_L1_stix-sci-cpd*.fits files), the shape of the counts array does not match the expected one according to the detector mask. This happens for 286 files (see list attached).
This produces an error when reading the file with procedure `stx_read_pixel_data_fits_file` in STIX-GSW.

In addition, in 14 cases also the number of energy bins doesn't match the shape of the counts array; but these are all from May 2020 (i.e. during NECP) and probably not an issue anymore.

The number of pixels according to pixel mask is very often less than 12 (in about 3,000 files), but this doesn't seem to be an issue for the file reading procedure, which always expects 12 pixels per detector, in agreement with the shape of the data array.

[stx_cpd_files_Ndet_mismatch.txt](https://github.com/i4Ds/STIXCore/files/10030684/stx_cpd_files_Ndet_mismatch.txt)

samaloney:
So they way the software currently works only the energy dimension changes based on the request so the for compressed pixel data or l1 (not L1) the count data will always be `t x 32 x 12 x e` where `t` is the number of times and `e` is the number of requested energies.

I had a look at the requests in you file you attached and the vast majority are for either the 30 imaging detector or only the background detector we could update the software to treat detector and pixels in a similar way to energies but not sure how difficult this would be or it it worth in in terms of the amount of space we could save. In the case of single detector requests there is a lot of wasted space in the fits files as `t x 31 x 12 x e` of the `t x 32 x 12 x e` entires will be all zeros.

samaloney:
@grazwegian @paolomassa @FredSchuller we discussed this today and decided it would be a good idea to do the above but we need to coordinate with the IDL code.

This PR makes all pixel data count dimensions vary, similarly to how the energy dimension currently works, matching the  number of requested energies. Currently all pixel data have shape `t x 32 x 12 x  e` where `t` is the number of times and `e` is the number of energies. When we reprocess the data after this change pixel data will be `t x d x p x e` where `d` is the number of detectors and `p` is the number of pixel. This save some space when we only request large pixels but a huge amount (~97%) when only BKG detector data is requested. Note the detector and pixel masks do and will continue to have sizes of 32 and 12. I think the easiest way is handle this is the when the data is read in you create the a `t x 32 x 12 x  e` and then it in from the data in the fits file the way the rest of the software can stay the same, will waste some memory but prob not huge issue.

[...]

PR #293, opened 2022-11-30
Fix offset between data and times and durations

samaloney:
I'm not sure how useful this really is as what we really need to know is the requested time bin size. As long as the time bins are `>=` the requested time bin size should be safe of shift otherwise not.


samaloney:
Discussed today and decided we should do where we know we can (most spectrogram and flare data) and add a flag or comment when we can't but need to.

setup.cfg
```diff
@@ -12,7 +12,7 @@ long_description = file: README.rst
 include_package_data = True
 zip_safe = False
 packages = find:
-python_requires = >=3.7
+python_requires = >=3.8
```
samaloney, line 15: I think this is causing the issues with the tests

stixcore/products/product.py
```diff
@@ -38,6 +39,7 @@
     32: np.uint32,
     64: np.uint64
 }
+DURATION_FIX_DATE = datetime(2021, 12, 9, 13, 0, tzinfo=timezone.utc)
```
nicHoch, line 43: still needed?

stixcore/products/product.py
```diff
@@ -706,6 +708,36 @@ def from_level0(cls, l0product, parent=''):
...
+                    correct_triggers = l1.data['triggers'][1:, :]
+                    correct_trigger_errs = l1.data['triggers_err'][1:, :]
+                    correct_times = l1.data['timedel'][1:]
+                    correct_durations = l1.data['timedel'][1:]
+                    correct_rcr = l1.data['rcr'][1:]
+                    l1.data = l1.data[0:-1]
+                    l1.data['counts'] = correct_counts
+                    l1.data['counts_err'] = correct_count_errs
+                    l1.data['triggers'] = correct_triggers
+                    l1.data['triggers_err'] = correct_trigger_errs
```
nicHoch, line 736: this is a nice and compact fix i was afraid it would have to be implemented on many more places.

nicHoch (approved):
good

stixcore/products/product.py
samaloney: yea I think my plan was to replace the datetime below I will do that now so would only be created once and then referenced

samaloney:
I think the only aspect missing is adding a comment when we know it should be fixed but we can't but this will be a fair but of work as have to add comments to the products and IO

nicHoch:
there is a "header" property on products if it was open from fits
i am not sure at what level 
but you could test if the property is around  and if so you can add the comment fairly easy

samaloney:
I don't quiet follow, what do you meant add the comment?

stixcore/products/product.py
```diff
@@ -130,18 +135,21 @@ def __call__(self, *args, **kwargs):
...
                 file_path = Path(args[0])
-                header = fits.getheader(file_path)
+                pri_header = fits.getheader(file_path)
+
+                service_type = int(pri_header.get('stype')) if 'stype' in pri_header else 0
+                service_subtype = int(pri_header.get('sstype')) if 'sstype' in pri_header else 0
+                parent = pri_header.get('parent').split(';') if 'parent' in pri_header else ''
+                raw = pri_header.get('raw_file').split(';') if 'raw_file' in pri_header else ''
+                comment = [c for c in pri_header.get('comment', [])]
+                history = [h for h in pri_header.get('history', [])]
```
nicHoch, line 145: ahh it is also reading in 
this is nice

stixcore/products/product.py
```diff
@@ -704,7 +716,43 @@ def from_level0(cls, l0product, parent=''):
...
+                    l1.data['counts_err'] = correct_count_errs
+                    l1.data['triggers'] = correct_triggers
+                    l1.data['triggers_err'] = correct_trigger_errs
+                    l1.data['time'] = correct_times
+                    l1.data['timedel'] = correct_durations
+                    l1.data['rcr'] = correct_rcr
+                    l1.history.append('Time and count arrays were shifted to fix offset')
+                else:
+                    l1.comment.append('Time and count arrays offset not fixed'
+                                      ' as possibly summed on board')
```
nicHoch, line 755: i know about that but somehow i think other users might not could we add a link to some kind of online documentation / git issue/ dpdd that will provide more background
?

nicHoch (approved):
i see now why you complained about the inheritance

stixcore/products/product.py
samaloney: Yea I think we need a read me with the data release summarising the data and know issues.

samaloney:
Closes #291

PR #295, opened 2022-12-06
Feat faster dsummask fix

samaloney:
Much faster lookup saves about 30s per BSD pixel request

```python
%timeit np.where((control['time_stamp'].coarse > 659318490) & (control['time_stamp'].coarse < 668863556))
20.1 µs ± 3.41 µs per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

%timeit print(control['time_stamp'].coarse not in range(659318490, 668863556))
37.8 s ± 6.42 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
```


nicHoch (approved):
looking good

PR #298, opened 2023-01-09
Update azure-pipelines.yml

samaloney:


stixcore/time/datetime.py
```diff
@@ -408,8 +408,8 @@ def from_float(cls, scet_float):
 
     @classmethod
     def from_btime(cls, btime):
-        coarse = btime >> np.int(16)
-        fine = btime - (coarse << np.int(16))
+        coarse = btime >> np.int_(16)
+        fine = btime - (coarse << np.int_(16))
```
samaloney, line 412: I don't really think this is need as `16` would be a int anyway

nicHoch (approved):
looking fine as discussed

PR #300, opened 2023-01-11
Update common config data files

samaloney:
Fix sci energy channel file


nicHoch (approved):
this will be interesting so see if we set up all update of submodules right also for the e2e test

i assume the changes of energy values in the e2e test are expected because of the new/modified LUT

PR #310, opened 2023-01-26
add quailty column to SAS (auxiliary fits)

nicHoch:


stixcore/products/level2/housekeepingL2.py
```diff
@@ -98,6 +98,7 @@ def from_level1(cls, l1product, parent='', idlprocessor=None):
             data['y_srf'] = 0.0
             data['z_srf'] = 0.0
             data['calib'] = 0.0
+            data['quality'] = 3
```
samaloney, line 101: Why 3, doesn't make much a of a difference anyway

samaloney (approved):

stixcore/products/level2/housekeepingL2.py
FredSchuller: Actually the default value should be 0, which means 'no problem'.
Maybe rename it 'badness' instead of 'quality', since my proposal (see [this issue in STIX-GSW](https://github.com/i4Ds/STIX-GSW/issues/129) was that higher values mean less reliable.

nicHoch: data['quality'] = 3 is the input/default value i pass into your code. so you should change that code to something less 3 if you know that your solution is good. (actually you should also resetit to 3 if you know your solution was bad/error).
it is set to 3 initially because if any connection error to idl happens then it returns a "true state": that there was an error udn you shoudl not use the SAS.

FredSchuller (approved):

PR #312, opened 2023-01-31
rename product descriptor from SPD to SCPD in L0

nicHoch:


stixcore/products/level0/scienceL0.py
```diff
@@ -492,7 +492,7 @@ class SummedPixelData(CompressedPixelData):
     def __init__(self, *, service_type, service_subtype, ssid, control, data, **kwargs):
         super().__init__(service_type=service_type, service_subtype=service_subtype,
                          ssid=ssid, control=control, data=data, **kwargs)
-        self.name = 'xray-spd'
+        self.name = 'xray-scpd'
```
samaloney, line 495: I don't know if we shouldn't change the class to match e.g. `SummedCompressPixelData`

samaloney (approved):

PR #313, opened 2023-01-31
provide none for soop keywords if no data

nicHoch:


stixcore/io/fits/processors.py
```diff
@@ -701,12 +701,12 @@ def generate_primary_header(self, filename, product, *, version=1):
                                                           otype=SoopObservationType.ALL)
         soop_headers = tuple(kw.tuple for kw in soop_keywords)
         soop_defaults = (
-            ('OBS_MODE', '', 'Observation mode'),
-            ('OBS_TYPE', '', 'Encoded version of OBS_MODE'),
-            ('OBS_ID', '', 'Unique ID of the individual observation'),
-            ('SOOPNAME', '', 'Name of the SOOP Campaign that the data belong to'),
-            ('SOOPTYPE', '', 'Campaign ID(s) that the data belong to'),
-            ('TARGET', '', 'Type of target from planning')
+            ('OBS_MODE', 'none', 'Observation mode'),
```
samaloney, line 704: I don't think `'none'` is a good idea as it could be interpreted as there was none rather than the value being empty or missing?

samaloney (approved):

PR #315, opened 2023-02-03
add soop api as input for all instruments soops

nicHoch:
added the api reader

i do not keep a reference to the api (coockies ...) as the in the processing pipeline the soop manager will stay alive many month 
so each time a file is updated a new connection (with auth) to the api is build

and handled the new version of plans

@samaloney please not that the API mocking is not done as it could be. But as we initialize the soop manager as a singleton  and it is referenced in many test cases without really testing the soop manager itself it is hard to but a mocking context around the manager


samaloney (approved):
Looks good to me, could maybe do with some more docstrings but not critical

PR #317, opened 2023-02-09
Fix and update various issues with BSD processing

samaloney:
* No longer rebin onboard summed data to standard science channels
* Fix count array energy dim not being truncated when onboard summing was used
* Rplace energy_bin_mask with energy_bin_edge_mask as the former is degenerate when onboard summing is used
* Update engery tables in fits to represnt channels in the data as requested

### DIFF
--- a/stixcore/products/level0/scienceL0.py
+++ b/stixcore/products/level0/scienceL0.py
@@ -5,13 +5,12 @@ import numpy as np
 
 import astropy.units as u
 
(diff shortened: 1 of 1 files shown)


samaloney:
e2e test changes look ok but hard be 100% sure

nicHoch (approved):
the goal of the PR sounds good but hard to tell just by looking on the code

PR #326, opened 2023-02-16
Fix regressions in cpd and spec

samaloney:
* Fix rare edge case with DSumMask fix and requested less than 32 detectors
* Fix spec count being effectively sliced along the energy dimension twice


nicHoch (commented):
ok

nicHoch (approved):
ok

PR #296, opened 2022-12-06
Feat save only requested detector and pixels to fits

samaloney:
When investigating #291 it was clear for some specific requests the files are 31 time larger than they need to be (when only the BKG detector) is requested. Further as many of our pixel data request are for large pixels only the are 1/3 larger than necessary. 

This PR subscrits the count array in the detector, pixel and engery dimensions so the files are as small as possible.

**This change would require updates to the python and IDL analysis software**

PR #328, opened 2023-02-20
Fix energy channel dimension of the RPD not being sliced back to requested energy channels

samaloney:
Forgot to do this for RPD when doing the the other data types in #296


samaloney:
Only need to reprocess rpd data.

nicHoch (approved):
looking good

PR #332, opened 2023-03-13
Leave float formatting to fits io

samaloney:
Currently force all spice fits header keywords to have a precision of 2 which is ok for values in m or km but not appropriate for values in degrees or even seconds this PR leave the formatting to fits io.


samaloney:
If you look at the e2e test all the values now have roughly the same overall length (including both digits before and after the decimal place)

nicHoch (approved):
good to go

Issue #333, opened 2023-03-16
Implement daily publishing file number limits to SOAR


nicHoch:
the GFTS server can only handle a certain number of new files per day.

the upper limit is 40.000

for the initial data load but also after possible reprocessing we would easily have more files to upload.

Implement a configurable daily (per run) limit of maximum files to publish

nicHoch:
closed in https://github.com/i4Ds/STIXCore/pull/334

PR #334, opened 2023-03-21
add batch_size limit for max # of files to publish at one (dayli) run

nicHoch:
addresses #333


samaloney (commented):
Looks good. So the idea is publish could be run a a number of times and at then end hopefully all the files will be published?

PR #335, opened 2023-03-28
HK aspect comments

nicHoch:
adding columns description but also dpdd notes that the aspect diode voltages are aggregated  

end2end test failed as expected as a new comment is stored in the extension header and everything is shifted afterwards.

also updates the FILENAME keyword in case of any sub1/2 while publishing


samaloney (commented):
Looks good

PR #336, opened 2023-03-29
Fix time shift corrections code

samaloney:


nicHoch (approved):
less code is better ;-)

PR #356, opened 2023-07-26
Handle cases where SPICE pointing information is not available

samaloney:
Add try catch around rotation matrix calculation and if not available set roll, yaw and pitch to nans.

### DIFF
--- a/stixcore/ephemeris/manager.py
+++ b/stixcore/ephemeris/manager.py
@@ -300,9 +300,14 @@ class Spice(SpiceKernelLoader, metaclass=Singleton):
         et = spiceypy.scs2e(SOLAR_ORBITER_ID, str(date))
         sc = spiceypy.sce2c(SOLAR_ORBITER_ID, et)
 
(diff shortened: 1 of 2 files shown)


samaloney:
Test fails seem unrelated.

nicHoch (commented):
this is looking good.
even if we now handle the most common error i wonder if it would be also good to general add an exception handler in within the batch loop at: 

https://github.com/i4Ds/STIXCore/blob/6dd81930999b971ae4606fa9226deaefbe3361e1/stixcore/products/level2/housekeepingL2.py#L256

the idl is called for batches of files.
any error in a file will terminate all files in the batch

samaloney:
Around this line
https://github.com/i4Ds/STIXCore/blob/6dd81930999b971ae4606fa9226deaefbe3361e1/stixcore/products/level2/housekeepingL2.py#L289 or in general? We don't really handle other kinds of missing spice information so I'm not sure.

nicHoch:
no the for each "file" loop here: https://github.com/i4Ds/STIXCore/blob/6dd81930999b971ae4606fa9226deaefbe3361e1/stixcore/products/level2/housekeepingL2.py#L256

but this is more general failure handle. so i can do it also in a dedicated PR

nicHoch (approved):

PR #360, opened 2023-09-21
Fix bug when adding HK products which use full SCET time.

### LINKED ISSUE #357: Wrong number of rows in L1 HK and L2 AUX files
I found in some L2 AUX files far too many rows, because some measurements appear twice (with time delays of less than 0.01s) - see examples in attached plots. The problem is already present in the L1 hk-maxi files, see for example 2023-03-29 and 2023-03-30. This is similar to issue https://github.com/i4Ds/STIXCore/issues/237 which was closed as solved a while ago.

time_vs_row_number AUX_file_20230329
time_vs_row_number AUX_file_20230330

samaloney:
When adding two producst one of the time arrays comes from TM so has full SCET timestamp, the otehr comes from the fits file which has been converted to a relative time and rounded. Thus when compared they appear different and the merge keeps both. This commit rounds the time from TM in simlar manner to when written to fits.

Closes #357

### DIFF
--- a/stixcore/products/product.py
+++ b/stixcore/products/product.py
@@ -525,7 +525,7 @@ class GenericProduct(BaseProduct):
 
         # Fits write we do np.around(time - start_time).as_float().to(u.cs)).astype("uint32"))
         # So need to do something similar here to avoid comparing un-rounded value to rounded values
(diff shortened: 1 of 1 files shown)


nicHoch (approved):
this is looking good.
nice one liner!

PR #372, opened 2023-11-14
add pipeline status monitor

nicHoch:
Status logger and notification script for the pipeline.

    SetUp via cron.
    Query the number of open files still to process. Logs that number into a status file
    and checks if the the number is constantly equal or increasing.

    Sends an notification via mail if a possible pipeline stuck is detected.


nicHoch:
@samaloney "unfortunately" we missed a stuck case on this weekend. i waited until i get the mail but forgot that i changed to another branch in me dev environment ...
but if you review it i can merge it into master soon

samaloney (approved):
If this is really a script maybe belongs in script folder rather than here?

nicHoch:
all our entry-points / mains clis for regular use are in this processing folder. the script are more used as helpers on times ...

i will also add a entry point in the project setup

Issue #339, opened 2023-04-03
Copy request ID into primary header where applicable


samaloney:
The request ID is one of the key parameters to identify a request currently it is stored in the control extension table would be useful to add as key in primary header too.

PR #373, opened 2023-11-14
add rid to FITS header

nicHoch:
address #339

### DIFF
--- a/stixcore/io/fits/processors.py
+++ b/stixcore/io/fits/processors.py
@@ -107,6 +107,18 @@ class FitsProcessor:
     @classmethod
     def generate_common_header(cls, filename, product, *, version=0):
 
+        user_req = ''
+        if 'request_id' in product.control.colnames:
+            rid_entry = np.atleast_1d(product.control['request_id'][0])
+            if len(rid_entry) > 1:
+                user_req = f'{rid_entry[1]:010d}-{rid_entry[0]:05d}'
+            else:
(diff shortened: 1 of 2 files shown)


nicHoch:
End2End Fails as expected as we have new keyword

grafik

stixcore/io/fits/processors.py
```diff
@@ -107,6 +107,18 @@ def generate_filename(cls, product, *, version, date_range, status='', header=Tr
     @classmethod
     def generate_common_header(cls, filename, product, *, version=0):
 
+        user_req = ''
+        if 'request_id' in product.control.colnames:
+            rid_entry = np.atleast_1d(product.control['request_id'][0])
+            if len(rid_entry) > 1:
+                user_req = f'{rid_entry[1]:010d}-{rid_entry[0]:05d}'
```
samaloney, line 114: Can/does this every happen what does `f'{rid_entry[1]:010d}-{rid_entry[0]:05d}'` look like?

nicHoch: yes this will happen in the cases where the rid is composed with RID and TC-SeqeunceID 

`solo_LB_stix-21-6-20_0000000000-9999999999_V01_0079424512-50488.fits`

in the `request_id` column we have an array in this case:

`array([   50488, 79424512])`

the entry value in the header will then also be: `0079424512-50488`

nicHoch:
@samaloney ready for a re-review

samaloney (approved):

PR #375, opened 2023-11-29
Various small fixes for data release

### LINKED ISSUE #338: Remove energy extension from calibration data
The energy extension is not applicable to calibration spectra datas so should be removed.

### LINKED ISSUE #355: Value of Carrington longitude is not correct
In the FITS header, the values given for "s/c heliographic longitude" (HGLN_OBS) and "s/c Carrington longitude" (CRLN_OBS) are identical, but they should not be.
For example, on 2022-04-16, the value of HGLN_OBS was around 148 deg, but CRLN_OBS should be around 343 deg.

### LINKED ISSUE #359: Slight discrepancies in energy edges
For the current fine thermal energy scheme the science energy channels are not at integer values. The numbers recorded in the FITS files are slightly different than expected e.g. the 6.56 keV edge is stored as 6.558594, while this is a very small discrepancy 0.02% it is higher than expected. 

Additionally these get magnified when taking the difference to calculate the bin width resulting in a ~1% discrepancy between the nominal value of 0.24 and the estimated value from the stored edges (0.242..).  

fine_energy_edges

### LINKED ISSUE #367: num_energy_groups changes rapidly with time
Screenshot 2023-10-24 at 15 32 23

When reviewing a issue related to the ELUT I noticed that the reported value in num_energy_groups seems to be different for each energy bin ranging between 3 and 15, I would have expected a constant value of 18 the total number of energy bins in for each data point. The DPDD gives the description "number of energy bands" which would seem to match my expectation. Checking with fv gives the same result suggesting this is not related to the IDL software. 

e.g. using the file from the imaging demo: 
fits_path  = solo_L1_stix-sci-xray-cpd_20200607T213709-20200607T215208_V01_1178428688-49155.fits 
data = stx_read_fits(fits_path, 'data')
data.num_energy_groups = [15  12   9   6   3  15  12   9   6   3  15  12   9   6   3  15  12   9   6   3  15  12   9   6   3  15  12   9   6   3  15  12   9   6   3  15  12   9   6   3  15  12   9  6   3]

samaloney:
Fixes:
1. Actually calculate the Carrington coordinate frame values and use in the headers, closes #355
2. Fix casting issue with fits ext for energy table, closes #359
3. Remove num_energy_groups, closes #367 
4. Remove energy extension from QL Calibration data, close #338

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -39,6 +39,7 @@ jobs:
     envs:
       - macos: py39
         name: py39_mac
+        coverage: false
 
       - windows: py39
         name: py39_win
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -35,6 +35,7 @@ jobs:
     toxdeps: tox-pypi-filter
     submodules: true
     coverage: codecov
+    default_python: '3.9'
     envs:
       - macos: py39
         name: py39_mac
--- a/setup.cfg
+++ b/setup.cfg
@@ -66,11 +66,15 @@ max-line-length = 100
 testpaths = "stixcore" "docs"
 doctest_plus = enabled
 text_file_format = rst
-addopts = --doctest-rst
+addopts =
+    --doctest-rst
+    --strict-markers
 remote_data_strict = False
 rsyncdirs = stixcore/data
 filterwarnings =
     ignore:Matplotlib is currently using agg:UserWarning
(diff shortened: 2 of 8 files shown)


samaloney:
E2E differences look as expected to me but 2nd look would be good.

nicHoch (approved):
this is looking good

PR #377, opened 2023-12-06
Add support for trigger scaling in compression mode S=0,k=0,M=7

### LINKED ISSUE #365: Support new scaling of triggers without compression.
The errors introduced by compressing the triggers is having a more adverse affect than anticipated so it is planned to scale the trigger to range 0 - 255. In this plan a new scheme a compression scheme `0,0,7`. Then based on the requested detector mask the triggers can be rescaled/unscaled back to their original values.

Meeting notes with code https://v000658.fhnw.ch/index.php/apps/dashboard/

- [x] Get scaling scheme
   * `Ns = No / (30 * Nacc * Nint)`
   * Where for example in TM(21, 6, 21) `Nint` ~ NIX00405 and `Nacc` ~ NIX00407 (converted to trigger groups)
   * An example from the AX log 202310152214-terminal.log gives `n_int=5`, `n_acc=1`, `No=1000` and `Ns=6` but evaluating that gives `1000/(30*5*1) = 6.66` so need to be careful about how it is handle on board floor, ceil or round.
   * Confirmed onboard FSW also essentially floors the value so to reconstruct need to add 0.5 to scaled value with an error of +/- 0.5 before un-scaling in the above example `6` -> `6.5` -> `6.5 * 30 * 5 * 1 = 975 +/-75` = `6 * 30 * 5 * 1 900` to `7 * 30 * 5 * 1 = 1050` 
- [ ] Get scaling parameters
   Will become parameter in FSW but fixed for no at `30`

samaloney:
Unscale trigger data when compressed with mode S=0,k=0,M=7

 * QL - LC, BKG, SPECTRA
 * BSD -  CPD, SCPD, VIS and SPEC

When compressed with mode 0,0,7 the trigger are scaled by the FSW according to `T_s = T / (n_int * n_acc * factor)` this scaling is removed by the GSW.

Closes #365

### DIFF
--- a/stixcore/products/common.py
+++ b/stixcore/products/common.py
@@ -321,20 +321,18 @@ def unscale_triggers(scaled_triggers, *, integration, detector_masks, ssid, fact
     active_trigger_groups = active_detectors_per_trigger_group >= 1
 
     # QL are summed to total trigger (1 trigger value)
-    if ssid in {30, 31, 32}:
+    # BSD SPEC data are summed to total trigger (1 trigger value)
+    if ssid in {30, 31, 32, 24}:
         n_group = active_trigger_groups.astype(int).sum()
+        n_int = integration.as_float().to_value(u.ds).reshape(-1, 1)  # units of 0.1s
     # BSD pixel/vis data not summed (16 trigger values)
     elif ssid in {21, 22, 23}:
         n_group = active_trigger_groups.astype(int)
-    # BSD SPEC data are summed to total trigger (1 trigger value)
-    elif ssid == 24:
-        n_group = active_trigger_groups.astype(int).sum()
+        n_int = integration.as_float().to_value(u.ds)  # units of 0.1s
     else:
         raise ValueError(f'Unscaling not support for SSID {ssid}')
 
-    n_int = integration.as_float().to_value(u.ds)  # units of 0.1s
-
-    # Scaled to ints onboard, so bins have scaled width of 1, so error is 0.5 times the total factor
+    # Scaled to ints onboard, bins have scaled width of 1, so error is 0.5 times the total factor
     scaling_error = np.full_like(scaled_triggers, 0.5, dtype=float) * n_group.T * n_int * factor
     # The FSW essential floors the value so add 0.5 so trigger is the centre of range +/- error
     unscaled_triggers = (scaled_triggers * n_group.T * n_int * factor) + scaling_error
--- a/stixcore/products/common.py
+++ b/stixcore/products/common.py
@@ -2,6 +2,8 @@ from pathlib import Path
 
 import numpy as np
 
+import astropy.units as u
+
 from stixcore.config.reader import get_sci_channels, read_energy_channels
 from stixcore.util.logging import get_logger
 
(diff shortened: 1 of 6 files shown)


samaloney:
This will need to be rebased on #375 once merged to fix the CI tests.

stixcore/products/common.py
```diff
@@ -278,3 +280,63 @@ def rebin_proportional(y1, x1, x2):
...
+    -------
+    tuple
+        Unscaled triggers, Scaling Variance
+    """
+    # TODO extract this to a separate function and test
+    detector_to_trigger_group_map = np.array(
+        [[1,  6,  5, 12, 14, 10,  8,  3, 31, 26, 22, 20, 18, 17, 24, 29],
+         [2,  7, 11, 13, 15, 16,  9,  4, 32, 27, 28, 21, 19, 23, 25, 30]]).T
+
+    trigger_groups = detector_masks[:, [np.array(detector_to_trigger_group_map) - 1]]
```
nicHoch, line 319: is this not defined in any other place CONF ...

stixcore/products/common.py
```diff
@@ -278,3 +280,63 @@ def rebin_proportional(y1, x1, x2):
...
+    # BSD SPEC data are summed to total trigger (1 trigger value)
+    elif ssid == 24:
+        n_group = active_trigger_groups.astype(int).sum()
+    else:
+        raise ValueError(f'Unscaling not support for SSID {ssid}')
+
+    n_int = integration.as_float().to_value(u.ds)  # units of 0.1s
+
+    # Scaled to ints onboard, so bins have scaled width of 1, so error is 0.5 times the total factor
+    scaling_error = np.full_like(scaled_triggers, 0.5, dtype=float) * n_group.T * n_int * factor
```
nicHoch, line 338: ok good point i would not have thought on that

stixcore/products/product.py
```diff
@@ -92,18 +92,72 @@ def read_qtable(file, hdu, hdul=None):
...
+        dtype :
+            The type to store the data as
+        reshape: `bool`
+            Reshape to a `1, X` array
+
+
+        Returns
+        -------
+        None
+        """
```
nicHoch, line 115: nice more doc is always good

stixcore/products/level0/quicklookL0.py
```diff
@@ -321,6 +332,13 @@ def from_levelb(cls, levelb, parent=''):
         counts_var = np.pad(counts_var, ((pad_before, pad_after), (0, 0)), constant_values=0)
         triggers = packets.get_value('NIX00484').T.reshape(-1)
         triggers_var = packets.get_value('NIX00484', attr='error').T.reshape(-1)
+
+        if control['compression_scheme_triggers_skm'].tolist == [[0, 0, 7]]:
+            logger.debug('Unscaling trigger ')
```
nicHoch, line 337:
```suggestion
        if control['compression_scheme_triggers_skm'].tolist() == [[0, 0, 7]]:
            logger.debug('Unscaling trigger ')
```

nicHoch: i did not run this but this looks odd

nicHoch (changes requested):
the tolist looks like an error to me.

at the moment it looks like the factor 30 is hard coded
the idea is to have a time base lookup somehow in the future?

stixcore/products/level0/quicklookL0.py
samaloney: Yea that was mistake fixed it now.

stixcore/products/common.py
samaloney: Yea it probably should be but at the moment this is the only place it's used if we merged the full conf from the STIX-GSW and STIX-CONF it would all be in one place.

samaloney:
Not to sure about the E2E test fails not related but wasn't expecting them either ...

nicHoch (approved):
good to go
was the test with the latest AX ql-spectra data successful?

samaloney:
Yea well I was able to produce fits files, I tried to compare the triggers from the QL lightcurve, background and spectra and they didn't match but hard to know if they should based on the different detector masks etc so until we get real data not much more can be done.

PR #379, opened 2024-01-30
Fix duplicated hk time bin bug

samaloney:


stixcore/products/product.py
```diff
@@ -580,21 +580,20 @@ def __add__(self, other):
         other_data['old_index'] = [f"o{i}" for i in other_data['control_index']]
         self_data['old_index'] = [f"s{i}" for i in self_data['control_index']]
 
-        control = vstack((self_control, other_control))
+        if (self.service_type, self.service_subtype) == (3, 25):
```
samaloney, line 583: I'm not sure if we should do the same for other TM types basically anything not (21, 6)?

stixcore/products/product.py
```diff
@@ -580,21 +580,20 @@ def __add__(self, other):
...
+            other_data['time'] = SCETime(other_control['scet_coarse'], other_control['scet_fine'])
 
         logger.debug('len self: %d, len other %d', len(self_data), len(other_data))
-
-        data = (vstack((self_data, other_data))
-                if self.scet_timerange.start <= other.scet_timerange.start
-                else vstack((other_data, self_data)))
-
+        control = vstack((self_control, other_control))
+        data = vstack((self_data, other_data))
```
samaloney, line 589: It's sorted later on line 589 so shouldn't matter which way it is stacked

stixcore/time/datetime.py
```diff
@@ -446,6 +452,15 @@ def to_datetime(self):
         except TypeError:
             utc = Spice.instance.scet_to_datetime(self.to_string())
 
+        kernel_name = Spice.instance.meta_kernel_path[1].name.split('_')[4]
```
samaloney, line 455: We really shouldn't use any time conversations in the future compared to the meta kernel date as they are subject to change especially for science data!

stixcore/products/product.py
nicHoch: lets see in future

stixcore/products/product.py
nicHoch: ok

nicHoch (approved):
also end2end looks good minor timing diffs in the right packages

good to go

PR #382, opened 2024-02-15
Fix bugs found when processing new scaled trigger data.

samaloney:
* Allow for QL files to unscale by packet
* Fix bug the replicate extra triggers in Spectrogram

### DIFF
--- a/stixcore/products/level0/quicklookL0.py
+++ b/stixcore/products/level0/quicklookL0.py
@@ -24,6 +24,9 @@ from stixcore.util.logging import get_logger
 
 __all__ = ['QLProduct', 'LightCurve', 'Background', 'Spectra']
 
+# SKM for scaling
+_SKM_SCALING = [0, 0, 7]
+
 logger = get_logger(__name__)
 
 QLNIX00405_off = 0.1
@@ -156,7 +159,7 @@ class LightCurve(QLProduct):
         cur_index = 0
         for i, (num, skm) in enumerate(zip(control['num_samples'],
                                            control['compression_scheme_triggers_skm'])):
-            if skm.tolist() == [0, 0, 7]:
+            if skm.tolist() == _SKM_SCALING:
                 cur_slice = slice(cur_index, cur_index+num)
                 logger.debug('Unscaling triggers')
                 triggers[cur_slice], triggers_var[cur_slice] = unscale_triggers(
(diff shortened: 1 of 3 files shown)


samaloney:
Hum I need to check some of the differences in the end2end tests ...

stixcore/products/level0/quicklookL0.py
```diff
@@ -241,17 +247,23 @@ def from_levelb(cls, levelb, parent=''):
...
-            logger.debug('Unscaling trigger ')
-            triggers, triggers_var = unscale_triggers(
-                triggers, integration=duration, detector_masks=dmask,
-                ssid=levelb.ssid)
+
+        cur_index = 0
+        for i, (num, skm) in enumerate(zip(control['num_samples'],
+                                           control['compression_scheme_triggers_skm'])):
+            if skm.tolist() == [0, 0, 7]:
+                cur_slice = slice(cur_index, cur_index+num)
```
nicHoch, line 261: maybe use of a constant like 
if skm.tolist() == SCALE_PATTERN

nicHoch (approved):
ok looks like a bigger fix 
but is looking good to me

samaloney:
Also I don't believe the test fail is related to git lfs as if you look at the checkout step from all the test they look very similar and seem to be doing the git-lfs checkout

stixcore/products/level0/quicklookL0.py
```diff
@@ -150,13 +153,19 @@ def from_levelb(cls, levelb, parent=''):
...
+        cur_index = 0
+        for i, (num, skm) in enumerate(zip(control['num_samples'],
+                                           control['compression_scheme_triggers_skm'])):
+            if skm.tolist() == _SKM_SCALING:
+                cur_slice = slice(cur_index, cur_index+num)
+                logger.debug('Unscaling triggers')
+                triggers[cur_slice], triggers_var[cur_slice] = unscale_triggers(
+                    triggers[cur_slice], integration=duration[cur_slice],
+                    detector_masks=control['detector_mask'][i:i+1], ssid=levelb.ssid)
+            cur_index += num
```
samaloney, line 168: Possible optimisation and could be extended for the other science data is rather than looping over all packets loop over data where skm apply usually only 1 skm and in the vast majority of cases number of different skm << num of packets.

PR #393, opened 2024-05-13
Flexible way to fill DATAMAX/MIN BUNIT EXPOSURE in FITS header keywords

nicHoch:
so far only products with "count" data fill in this keywords in a usefull way. with this more flexible way of generating it should be much better to fill this keywords for deticated products


stixcore/products/level2/housekeepingL2.py
```diff
@@ -342,6 +342,18 @@ def __init__(self, *, service_type=0, service_subtype=0, ssid=1, control, data,
         self.service_subtype = 0
         self.service_type = 0
 
+    @property
+    def dmin(self):
+        return min([self.data['y_srf'].min().value, self.data['z_srf'].min().value])
```
samaloney, line 347: Better to use `to_value('km')` incase the `value` is in different units, not the case here but doesn't hurt.

stixcore/products/level1/quicklookL1.py
```diff
@@ -109,6 +130,19 @@ def __init__(self, *, service_type, service_subtype, ssid, control, data,
         self.name = 'flareflag'
         self.level = 'L1'
 
+    @property
+    def dmin(self):
+        return min([self.data['loc_y'].min(), self.data['loc_z'].min()])
```
samaloney, line 135: Is `value` or `to_value()` not needed here as below?

stixcore/products/level1/scienceL1.py
```diff
@@ -75,6 +75,21 @@ def __init__(self, *, service_type, service_subtype, ssid, control,
         self.name = 'xray-vis'
         self.level = 'L1'
 
+    @property
+    def dmin(self):
```
samaloney, line 79: Yea this might be tricky as they are complex numbers

samaloney (approved):
LGTM 

In general always better to use `to_value(<unit>)` over `value` as if in different units value just give the raw number ignoring possible difference e.g unit prefix difference `km` v `m`.

stixcore/products/product.py
```diff
@@ -743,6 +768,28 @@ def fits_daily_file(self):
...
+    def dmax(self):
+        return self.data['counts'].max().value
+
+    @property
+    def bunit(self):
+        return 'counts'
+
+    @property
+    def exposure(self):
+        return self.data['timedel'].as_float().min().to_value('s')
```
samaloney, line 786: might resolve the test errors not sure
```suggestion
        return self.data['timedel'].min().to_value('s').as_float()
```

nicHoch: fixed

samaloney:
Hum most of the test fails look ok now but the ones L1530 and L2008 don't seem related and are unexpected at least at first look.

PR #411, opened 2024-11-11
Get rid reason/scaling lookup error handling

nicHoch:


samaloney (approved):
LGTM

Just to confirm the E2E test fails are due to the python issue on the server?

PR #423, opened 2025-05-13
Fix bug in spectrogram time bin calculations

samaloney:
* Only use the closing time offset (NIX00269) for the last science substructure to calculate the time bins.


samaloney:
Closes #424

nicHoch (approved):
all good

Issue #421, opened 2024-12-19
Error processing QL Spectra from LB to L0


samaloney:
Since the files are created in the end most likely happens due to partial data in one TM download and then causing the error but in next TM download get the rest of the data.

Example log file https://pub099.cs.technik.fhnw.ch/fits/logging/TM_hourly_req_BatchRequest.PktTmRaw.SOL.0.2024.339.09.53.50.363.CMpR@2024.354.13.00.03.935.1.xml.log.err

```
2024-12-19T18:04:19Z ERROR stixcore.processing.LBtoL0 95: Error processing file /data/stix/out/fits_v1.2.0/LB/21/6/32/solo_LB_stix-21-6-32_0787881600_V02.fits for 21, 6, 32
Traceback (most recent call last):
  File "/home/stixcore/STIXCore/venv/lib/python3.9/site-packages/stixcore/processing/LBtoL0.py", line 90, in process_tm_type
    level0 = tmp.from_levelb(levelb, parent=file.name)
  File "/home/stixcore/STIXCore/venv/lib/python3.9/site-packages/stixcore/products/level0/quicklookL0.py", line 344, in from_levelb
    control_indices = control_indices.reshape(-1, 32)
ValueError: cannot reshape array of size 1644 into shape (32)
2024-12-19T18:04:19Z ERROR stixcore.processing.LBtoL0 97: cannot reshape array of size 1644 into shape (32)
```

samaloney:
Closed via #426 at least for the moment

PR #426, opened 2025-07-15
Don't raise an error on QL spectra reshape emit warning instead

samaloney:
At the moment we get these error every day and may be missing real errors as in the end the QL spectra a created properly. 

This PR attempts to disable the error to see will help until we can reproduce and fix the error (#421).


nicHoch (approved):
all good

samaloney:
Looking at this again I'm not sure it won't just error at a later line ...

nicHoch (approved):
ready to go

PR #427, opened 2025-07-22
CAL spectra : daily energy calibration 

nicHoch:
this PR started a while ago as first version of general L2 products as QL calibration spectra and sci-spectra as L2

then it moved its goal to daily energy calibration files that have become an own "level": CAL

so some code parts for L2 files are commented out to be eneabled later again with our first L2 products


stixcore/soop/manager.py
```diff
@@ -529,6 +529,8 @@ def add_soop_file_to_index(self, path, *, rebuild_index=True, **args):
             all_soop_file = Path(CONFIG.get("SOOP", "soop_files_download")) / f"{plan}.{version}.all.json"
 
             if not all_soop_file.exists():
+                # TODO reactivate when API is back
+                return
```
samaloney: Don't forget to remove now

samaloney (approved):
The managers are all very similar I wonder if we shouldn't make a general manager and then have specific versions?

Also I think we should avoid adding more of the singleton pattern, I know I was the one who started with it but it really an anti-pattern in python in general.

PR #429, opened 2025-08-27
Mirgrate to use stixpy calibration

samaloney:
* Remove calibration module from STIXCore
* Change imports to use calibration from stixpy


nicHoch (approved):
this is good to go

PR #439, opened 2025-11-11
More update for numpy 2.0 casting changes in product __add__

samaloney:
When adding two products need to ensure the index `dtype` column is sufficient to hold the combine number of rows.


nicHoch (approved):
good to go

PR #440, opened 2025-11-14
V1.6.0_release_prep_fixes

nicHoch:
* fix missing ascpect class var for file name generation
* some more logging
* fix "memory leak" in packet parsing: very larhe packet sequences can lead to a OoM-Exception as each packets holds a reference to the packet parse tree (for debugging and printing) this strukture is arround 1MB for each packet. in the normal pipeline usage that structure should not be tracked as it is to memory extense and is not used. Added a default keyword to prune that data

### DIFF
--- a/stixcore/idb/tests/test_manager.py
+++ b/stixcore/idb/tests/test_manager.py
@@ -10,7 +10,6 @@ from stixcore.time import SCETime
 
 
 @pytest.fixture
-@pytest.mark.remote_data
 def idb_manager():
     return IDBManager(test_data.idb.DIR)
 


stixcore/processing/pipeline_cron.py
```diff
@@ -169,9 +168,9 @@ def __init__(self, tm_list):
         self.current_tm = (None, datetime.now())
         self.tm_list = tm_list
 
-        self.status_server_thread = threading.Thread(target=self.status_server)
-        self.status_server_thread.daemon = True
-        self.status_server_thread.start()
+        # self.status_server_thread = threading.Thread(target=self.status_server)
+        # self.status_server_thread.daemon = True
+        # self.status_server_thread.start()
```
samaloney: Same here?

samaloney (approved):
LGTM - might be good to remove some of the commented out code or leave comment explaining why it commented out?

Do we ever use the parse tree in normal processing?

PR #447, opened 2025-12-05
add option to fits reader to open L1 products also in UTC >L1 UTC per default (via keyword in header)

nicHoch:
add keyword: get_timeformat_from_TIMESYS to Product(path) method.
If set time and timedel column in data table are converted to timesystem provided in the  TIMESYS header keyword.

some products properties like utc_timerange, min_exposure ... might do autoconversions based on the internal timesystem but it should be avoided to open L1 files in UTC for processing them further. 

Use get_timeformat_from_TIMESYS=True only for reading excess to L1 data, read/write/combing might lead to timing accuracy issues

### DIFF
--- a/stixcore/processing/tests/test_end2end.py
+++ b/stixcore/processing/tests/test_end2end.py
@@ -73,8 +73,8 @@ def test_complete(orig_fits, current_fits):
         raise ValueError(f"{error_c} errors out of {len(orig_fits)}\nnumber of fits files differ")
 
 
+# @pytest.mark.end2end
 @pytest.mark.remote_data
-@pytest.mark.end2end
 def test_identical(orig_fits, current_fits):
     error_c = 0
     error_files = list()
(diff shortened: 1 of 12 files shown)


samaloney (commented):
I'm not sure allowing operations in different time systems is a good idea. A product should either have time in OBT or UTC and the conversion should only be done once in the `from_level0` or similar methods?

All products (>L0) will always have `obt_timerange` from products will have an `utc_timerange` but only if the time sys in header is `UTC` and not sure supporting on the fly conversion is a good idea because the spice kernel dependence.

This aren't blocker just thoughts

stixcore/processing/tests/test_end2end.py
```diff
@@ -73,8 +73,8 @@ def test_complete(orig_fits, current_fits):
         raise ValueError(f"{error_c} errors out of {len(orig_fits)}\nnumber of fits files differ")
 
 
+# @pytest.mark.end2end
 @pytest.mark.remote_data
-@pytest.mark.end2end
```
samaloney: ?

samaloney (approved):
Seems ok, one test marker commented out?

